PR #36, opened 2020-11-16
example applicability template

mcr:
add an appendix as per discussion in meeting today.
Probably the example should be retwigged (as it contains my obscure humour), and likely I've missed some of the detail that needs to be determined.


gselander:
I merge this and make updates to it on the master branch

Issue #90, opened 2021-04-08
Clarify "auxiliary data"


gselander:
It is too easy to confuse "application data" with "auxiliary data". The former is the data which is protected with keys obtained through the use of EDHOC, i.e. a straightforward application of EDHOC so the name "application data" is fairly intuitive. The latter is an extension point for use with security applications during EDHOC execution, like third party assisted authorization or certificate enrolment. 

Perhaps we should rename "auxiliary data" something with "extension"?

emanjon:
I think extension is a bad word choice as it is easy to think about TLS extensions, which this is not similar to at all.

Maybe "external data"

emanjon:
I suggest "external authorization data" EAD. I think it is good to point out that the data is external to EDHOC and to give a hint of what it is used for.

emanjon:
Doing a pull request for this, I stumbled upon several things:

? AD_1 : bstr,

Why is this bstr instead of any. If the data is CBOR encoded in the first place, wrapping it in a bstr adds two bytes without any benefit I suggest we change this to

? AD_1 : any,

emanjon:
Specifying bstr also makes it harder to put any structure to the data in a formal way later as was discussed.

emanjon:
>The Responder MUST finish the verification step of message_3 before passing AD_3 to the application.

Why is this stated? I don't understand at all. The reason for AD_3 to exist is to send things like access tokens that are processed before message_3 is completely processed.

The current text makes it sound like AD_3 is application data which the agreement is that it is not.....

emanjon:
I made a PR #111 addressing all the issue mentioned with aux data.

The PR addresses issue #110

gselander:
Any comments on #111?

emanjon:
An earlier version of EDHOC stated that was then called EXT and is now called EAD "could be used for authorization tokens that are protected outside of EDHOC." The intention with that text was CWT. The text has since been removed without being being replaced with any other way to transfer CWT. CWT can in the current version not be transfered in EAD_2 or EAD_3

gselander:
Why can't CWT can be transferred in EAD_2 or EAD_3? As I recall the statement about "protected outside of EDHOC" that allowed protection within as well as outside of EDHOC.

emanjon:
It can, but it creates more overhead. Treating CWT differently than X509 seems a bit strange, especially in the opportunistic RPK case.

[...]

PR #111, opened 2021-05-04
"external authorization data" EAD

emanjon:


emanjon:
The definition says array but the CDDL is a CBOR sequence. In this case it works to use a sequence.

The CDDL for message_1, 2 3 4 needs to be updated. Right now they say 

? EAD_1 : any,

emanjon:
Might also want to allow more that one CBOR item. The following would be quite general:

EAD = (
  type : int,
  + ext_authz_data,
)

? EAD_1 : EAD,

Should probably write that "+ ext_authz_data," follows from "type : int,"







Extensions = [ * Extension ] / int

Extension = ( extensionID: int, extensionValue: any ) //
            ( extensionID: ~oid, critical: bool, extensionValue: bytes )
 )

marco-tiloca-sics:
Looks good.

emanjon:
Was there any reason that this was not merged today. The new IANA registry for Aux Data should probably be another PR.

gselander:
The reason for not merging this yet was to first conclude the discussion about CWT in EAD_2 and EAD_3 in #90.

gselander:
By moving out the IANA type to a separate issue #126 and waiting with the CWT related part of #90 we are ready to merge.

PR #113, opened 2021-05-05
Error codes and definitions.

emanjon:
This should address all open issues regarding errors.
- Align with very common standard that 0 means success and everything else means error
- Specify that success is for internal use in the application, it therefore makes sense to use any and let the application define what to store.
- No sense in forcign application to support 0, it is not needed for interoperability over the wire.


gselander:
I kept error codes positive as proposed by IOTOPS, but changed the more general error "unspecified" comes before the more specific "wrong ciphersuite" because 
1. that allows for more specific errors to follow in a structured way in the table
2. that aligned better with the subsubsections describing the cipher suite negotiation.

gselander:
John and I are happy with this now. Unless there are any comments we will soon merge this.

marco-tiloca-sics:
Looks good.

Issue #97, opened 2021-04-19
Forbidding multiple calls to EDHOC-Exporter with the same label


malishav:
The spec should likely forbid multiple calls to the EDHOC-Exporter interface with the same label, in order to prevent the same key being reused.

Discussed with @karthikbhargavan and @TimothyClaeys.

emanjon:
Good catch. Yes, we should have some text about that, but it should probably more be a requirement on the application using EDHOC that on EDHOC itself. Two different application should defintly not use the same label. The same application might rederive the same key several times as long as it keep track of its AEAD nonces and replay window.

- I thought that we had already IANA registry proposed for this, but that is not the case. We should probably have one.
- If we have a IANA registry the labels get more fixed and it might be that we need a context input parameter as well.

emanjon:
If we make a IANA registrer for labels, I think we need to let the application add additional information in some way. Two alternatives:

1. We make an IANA register for label prefixes, and let the application append information to the label prefix. This was recently planned for EAP-TLS (and methods built on EAP-TLS) but was changes because the TLS export labels are not really ment to be labels. This would not change the current exporter interface.

2. We expand the interface with a context parameter. This is similar to the TLS exporter. A problem with the TLS exporter is that some implementations decided to not even support context.

malishav:
@emanjon how is the context parameter intended to be used? Could you give an example? Would this be some sort of "salt" when deriving the key?

cabo:
Don't register labels, register label prefixes.

emanjon:
The motivation for "context" or "label postfixes" is to increase the flexibility for the application. Restricting the application to IANA registered labels is very limiting. An application might want to use the EDHOC exporter to derive a number of keys and we should encourage that. The alternative is that the application needs to implement its own key derivation, which might be less secure and might be outside of the TEE. I called it "context" just because TLS calls the paramater in its exporter context.

cabo:
"Be like TLS" is of course a good motivation.
Technically, a label prefix is not different from a context, it just saves the need to separate prefix and suffix.

cabo:
(And of course there is nothing wrong with multiple calls with the same label; the application has a "right to forget" :-)

emanjon:
The solutions should be equal when it comes to security. The choice should be made based on what is easiest for implementators and apllications using EDHOC.

Alt 1.
--------------------------

We make an IANA register for label prefixes:

  "OSCORE Master Secret"
  "OSCORE Master Salt"

[...]

PR #114, opened 2021-05-06
Exporter labels, IANA, context, requirements

emanjon:


marco-tiloca-sics:
The prototype of EDHOC-Exporter() has remained EDHOC-Exporter(label, length), while it should be EDHOC-Exporter(label, context, length) as per Alternative 3 discussed in issue https://github.com/lake-wg/edhoc/issues/97. Correct?

gselander:
Thanks @marco-tiloca-sics, your comment is addressed in the latest commit. I merge this now.

PR #117, opened 2021-05-17
Move the prepended C_x from the messages to CoAP section

chrysn:
... and add an appendix telling other protocol binding authors to do as
CoAP does.

This is an initial rough change, primarily guided by search for a few
keywords, so some spots that need changed wording may have been missed.

The test vectors were left unmodified. Some data structurs were modified
minimally to be now empty; removing them completely can only be done
informed my their relevance to the transcripts.

### DIFF
--- a/draft-ietf-lake-edhoc.md
+++ b/draft-ietf-lake-edhoc.md
@@ -265,21 +265,21 @@ To simplify for implementors, the use of CBOR and COSE in EDHOC is summarized in
 
 An EDHOC message flow consists of three mandatory messages (message_1, message_2, message_3) between Initiator and Responder, an optional fourth message (message_4), plus an EDHOC error message. EDHOC messages are CBOR Sequences {{RFC8742}}, see {{fig-flow}}. The protocol elements in the figure are introduced in the following sections. Message formatting and processing is specified in {{asym}} and {{error}}. An implementation may support only Initiator or only Responder. 
 
-Application data is protected using the agreed application algorithms (AEAD, hash) in the selected cipher suite (see {{cs}}) and the application can make use of the established connection identifiers C_1, C_I, and C_R (see {{corr}}). EDHOC may be used with the media type application/edhoc defined in {{iana}}.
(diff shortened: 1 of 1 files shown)


chrysn:
The stats (which seem to indicate that more text was added than removed) are skewed a bit by my aversion to long lines, the change log, and a few lines explaining why some empty data structures were left around for further editing. Compensating for those, the text was shortened by 11 lines or 116 words, and can possibly be shortened a bit more now that data_3 and data_4 are empty and might go away completely.

chrysn:
Spelling out the changes from this PR a bit more explicitly, the messages now all have zeros in the place where they used to have corr (METHOD may be shifted down to by two bits, not done here yet), and crypto-operation-wise behave as if it were 3 (as data_n now do not contain their leading C_x any more even in the client-to-server direction).

For messages 3 and 4 this means that these are now always empty. Right now I left them as empty arrays, but for data_3 that means that one hashing step is really just a double hashing (`TH_3 = H( H(TH_2, CIPHERTEXT_2), data_3 )` where data_3 is always the empty bytestring) -- can that hashing operation be removed?

malishav:
I am having a hard time understanding how this change is intended to be implemented: EDHOC messages still include the mandatory connections identifiers; optional identifiers are moved and handled by the transport. Part of the logic seems to be within the EDHOC machinery, part of the logic within the transport handler. @chrysn please let me know if I am missing something.

gselander:
@malishav Some reflections: 

> EDHOC messages still include the mandatory connections identifiers

EDHOC messages still include the negotiation of connections identifiers but strictly speaking they are not used by EDHOC anymore.

> Part of the logic seems to be within the EDHOC machinery, part of the logic within the transport handler.

Already with the previous version, in the case of CoAP, part of the logic was outside the EDHOC machinery and in the transport handler, like the use of the CoAP Token. Also, in case of OSCORE, part of the logic was outside the EDHOC machinery because EDHOC could not unilaterally decide on the values of connection identifiers without considering existing OSCORE Recipient Identifiers.

Is this really so different?

gselander:
See #129

malishav:
I am not arguing that this is impossible to implement, I am arguing that it is not the cleanest design. See my comment on #129  on how I think we could cleanly describe the separation of concerns.

emanjon:

Issue #79, opened 2021-03-01
Coding density for bstr_identifier


chrysn:
We could increase (by a factor of about 3) the number of identifier values encodable in 2-byte bstr_identifier -- if rather than treating them as a byte string toward the application, we treat them as distinct integer and bytes values.

As before, the same 1-byte values could be encoded (-24 to 23). Of the 2-byte long encodings, there there would be ~3x256 inhabited ones (unlike now, ~256): We could use the 256 `18 xx` (for xx > 24 as those would be encoded in a single byte), `38 xx` (again, xx < 24 staying uninhabited) and the `41 xx` (1-long byte strings) values.

Whether this makes implementation harder or easier is IMO dependent on the programming style; for me it'd make things easier. (Implementers can't store the value as a length-delimited bytestring with a custom encoder any more, but can still store the CBOR encoded item in an opaque string).

This would also allow applications that find it easier to work with small numbers to just use numeric identifiers for everything they mint themselves, and use these in lookup. (These then lose some of the benefits of identifier compression.)

chrysn:
Considering how the OSCORE extractor uses C_I and C_R, this would be problematic.

Options are:
1. Leave it as is -- with only ~256 2-byte C_x available
2. Allow arbitrary int / bstr as bstr_identifier, but move the text over to the OSCORE extraction to say that this mapping is applied between C_x and `kid`. That would leave the general applications with the wide possibilities, while OSCORE would have the 0-length KID in a 1-byte bstd_identifier, a few selected KIDs (00 to 18) also in the 1-byte bstr_identifier, and all the others in len+1 bstr_identifier slots.
3. Go for even more efficiency (but that has minimal returns at great complexity so is almost certainly not worth it)

I have a soft preference for 2, but 1 is probably fine as well -- given that OSCORE will be a main user of EDHOC and couldn't use the optimization.

emanjon:
I think 2 is the right way to go. I think we should do that. I think the current specification of bstr_identifier is a bit complex. We have gotten several comment and questions regarding this. I think 2 make things simpler.

emanjon:
I made a pull request for this issue. @chrysn check that the transformation between C_x and OSCORE ID is what you wanted. I did not see any point in making it more complex.

I just realized that we could expand the number of 1 byte excodings with almost 50 % by also allowing major type seven I.e.

C_x : bstr / int / #7 

(null is reserved for used by C_1)

emanjon:
[...]

Issue #75, opened 2021-02-21
Move 7.2.  Transferring EDHOC in CoAP to draft-palombini?


emanjon:
7.2 was written before draft-palombini existed. Would it make sense to move all OSCORE specific informaiton to draft-palombini and make that draft more general. Right now you probably have to read both. A lot of the new issues and discussion are concerning the OSCORE EDHOC interaction.

gselander:
As there has been no comment on this issue, we now plan for moving out section 7.2.1 and second half of 7.2.2 to draft-ietf-core-oscore-edhoc (former draft-palombini- ...)

StefanHri:
This makes sense for me.

marco-tiloca-sics:
Yes, that makes sense.

rikard-sics:
Sounds good to me also.

malishav:
As discussed during the last interim, I do not believe removing the link of draft-ietf-lake-edhoc with OSCORE is a good idea from the charter point of view of the LAKE WG.

I just went through the editor's version of draft-ietf-core-oscore-edhoc and here is my proposal forward:

* Create a new appendix in draft-ietf-lake-edhoc named "Use with OSCORE", say Appendix A.
* Move the content of Section 4 of draft-ietf-core-oscore-edhoc back to draft-ietf-lake-edhoc, to Appendix A.1.
* Merge the content of Section 3 of draft-ietf-core-oscore-edhoc with the content in Section 7.2 of draft-ietf-lake-edhoc. Move the result to Appendix A.2.

With that, draft-ietf-lake-edhoc contains all the information to implement EDHOC and use it to key OSCORE. draft-ietf-core-oscore-edhoc becomes streamlined and focuses on optimizing the specific use case when message_3 and OSCORE request are sent in the same CoAP message for performance reasons.

gselander:
@malishav @mcr I believe -08 does exactly what you requested. Can we close this?

malishav:
WFM, closing

PR #122, opened 2021-05-19
Drop bstr_identifier, allow bstr / int in C_x instead

chrysn:
The largest fall-out of this will be in draft-ietf-core-oscore-edhoc
since https://github.com/lake-wg/edhoc/issues/75.

Close: #79

---

As with #117, the test vectors are untouched. (I don't expect actual change in the numbers, just in the text).

This is incomplete on its own due to the #75 changes; I'm opening a separate PR there.

Another piece of incompleteness is that this references (with a remark) the kid-int variant suggested in #79, without actually defining it. (But that would be merely an IANA item; the big question to COSE being whether we can up-spec kid to bstr / int in-place or really need kid-int).

### DIFF
--- a/draft-ietf-lake-edhoc.md
+++ b/draft-ietf-lake-edhoc.md
@@ -312,7 +312,7 @@ An implementation may support only a single method. The Initiator and the Respon
 
 ### Connection Identifiers {#ci}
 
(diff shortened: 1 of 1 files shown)


gselander:
See also #98

emanjon:

PR #123, opened 2021-05-20
Simplify MAC calculation

emanjon:


marco-tiloca-sics:
Considering this PR on the MAC computation and the recent update to the Exporter interface, the EDHOC-KDF() interface in Section 4 "Key Derivation" now generally takes a CBOR sequence as third argument "label".

If I haven't missed anything in other PRs or commits, this should be reflected also in the definition of the 'info' array in that same section, which is still specifying its "label" element as a text string. Correct?

emanjon:
Replaces by #136 and #137

PR #127, opened 2021-06-01
cipher suites

emanjon:


gselander:
Looks good. One inconsistency is that there is no cipher suite with AES-GCM and NIST curves, like (A128GCM, SHA-256, P-256, ES256, A128GCM, SHA-256).

This could be a new cipher suite 5. Then all non-CNSA suites would come in pairs.

emanjon:
> Looks good. One inconsistency is that there is no cipher suite with AES-GCM and NIST curves, like (A128GCM, SHA-256, P-256, ES256, A128GCM, SHA-256).
> 
> This could be a new cipher suite 5. Then all non-CNSA suites would come in pairs.

That is very intentional. The GCM cipher suite uses X25519 for key exchange and ECDSA for authentication. This seems to be the dominant choice for HTTPS on the Web. Unless you are constrained you typically use X25519 for key exchange for speed and ECDSA for authentication as that is what the CAs support.

For an IoT device you probably want to have a single curve implemented. Also it might be much easier to switch to EdDSA as you might not have to rely on Global CAs and all other HTTPS nodes to support EdDSA. You can just deploy you own CWT server and make sure the nodes in you new IoT deployment support EdDSA.

I have not heard any IoT people very interested in GCM. People seems far more interested in ChaCha20.

emanjon:
Ready to merge?

PR #135, opened 2021-07-26
Update draft-ietf-lake-edhoc.md

emanjon:


emanjon:
Replaces by #136 and #137

PR #136, opened 2021-07-26
Simpler more secure MAC calculation (CBOR sequence)

emanjon:


chrysn:
At first glance, I like this better than #137, as here I don't have to keep the number of elements in EAD_3 around. (I'll have to have processed them at some point, but that doesn't mean I'm keeping the count around).

emanjon:
Based on Christians comment. I would suggest that we merge this. I think allowing a CBOR sequence in the exporter migth also be a good idea. That allows external applications to input a CBOR sequence. The implementation allows a implementation to input tsrt, bsrt. That is just a special case when the CBOR sequence consist of a single bstr.

draft-ietf-lake-edhoc.md
```diff
@@ -826,31 +829,14 @@ The Responder SHALL compose message_2 as follows:
 
 * Compute the transcript hash TH_2 = H( H(message_1), data_2 ) where H() is the hash function in the selected cipher suite. The transcript hash TH_2 is a CBOR encoded bstr and the input to the hash function is a CBOR Sequence. Note that H(message_1) can be computed and cached already in the processing of message_1.
 
-* Compute an inner COSE_Encrypt0 as defined in Section 5.3 of {{I-D.ietf-cose-rfc8152bis-struct}}, with the EDHOC AEAD algorithm in the selected cipher suite, K_2m, IV_2m, and the following parameters:
-
-   * protected =  << ID_CRED_R >>
+* Compute MAC_2 = EDHOC-KDF( PRK_3x2m, TH_2, ( "MAC_2", ID_CRED_R, CRED_R, ? AD_2 ), mac_length ). If the Responder authenticates with a static Diffie-Hellman key (method equals 1 or 3), then mac_length is equal to the tag length of the EDHOC AEAD algorithm. If the Responder authenticates with a signature key (method equals 0 or 2), then mac_length is equal to the output size of the EDHOC hash algorithm.
```
marco-tiloca-sics, line 832: Here the PRK should be PRK_3e2m.

draft-ietf-lake-edhoc.md
```diff
@@ -915,31 +901,14 @@ The Initiator SHALL compose message_3 as follows:
 
 * Compute the transcript hash TH_3 = H(TH_2, CIPHERTEXT_2) where H() is the hash function in the selected cipher suite. The transcript hash TH_3 is a CBOR encoded bstr and the input to the hash function is a CBOR Sequence.  Note that H(TH_2, CIPHERTEXT_2) can be computed and cached already in the processing of message_2.
 
-* Compute an inner COSE_Encrypt0 as defined in Section 5.3 of {{I-D.ietf-cose-rfc8152bis-struct}}, with the EDHOC AEAD algorithm in the selected cipher suite, K_3m, IV_3m, and the following parameters:
-
-   * protected =  << ID_CRED_I >>
+* Compute MAC_3 = EDHOC-KDF( PRK_4x3m, TH_3, ( "MAC_3", ID_CRED_I, CRED_I, ? AD_3 ), mac_length ). If the Initiator authenticates with a static Diffie-Hellman key (method equals 1 or 3), then mac_length is equal to the tag length of the EDHOC AEAD algorithm. If the Initiator authenticates with a signature key (method equals 0 or 2), then mac_length is equal to the output size of the EDHOC hash algorithm.
```
marco-tiloca-sics, line 904: It should be:

If the Initiator authenticates with a static Diffie-Hellman key (method equals 2 or 3) ... If the Initiator authenticates with a signature key (method equals 0 or 1)

draft-ietf-lake-edhoc.md
emanjon: fixed

draft-ietf-lake-edhoc.md
emanjon: Fixed

gselander (approved):

PR #137, opened 2021-07-26
Simpler more secure MAC calculation (CBOR array)

emanjon:


emanjon:
This is now a change bases on #136. The reader need to first look at #136 to understand the suggested changes to the main branch.

The diff gives a good overview of the difference between a sequence solution and a array solution.

If approved. #137 needs to be merged before #136.

emanjon:
Closing this, the plan is to merge #136

Issue #103, opened 2021-04-22
Optimization of message size


gselander:
We need to nail down what optimizations to make in the final message format. 

Known lengths of byte strings can be used as discussed in the old issue tracker:
https://github.com/EricssonResearch/EDHOC/issues/95

emanjon:
We should confirm the values from the rewuirement phase. If I remember correctly 45 bytes for message_2 in 5-hop 6TiSCH was the strictest requirement. We can reasch that but before we optimize for that, we should confirm that the group wants to optimize for 5-hop 6TiSCH (or is 4-hop godd enough) and also double check and verify that 45 bytes is the correct number to optimize for.

If I remember correctly Lorawan was a little bit less demanding than 5-hop 6TiSCH wih 48 bytes?.

Every single byte can make a large difference if you suddenly have to fragment and send two frames instead of one. On the other hand there has been discusssions that is would be nice to mandate C_R in message_2 and to skip the bstr_identifier optimization of ID_CRED_R.

gselander:
John added a comment on an issue in the old repo which probably was intended to be on this issue: 
https://github.com/EricssonResearch/EDHOC/issues/95

emanjon:
Several things have been discussed regarding optimisations (or simplifications adding bytes).

Message_2 is the largest message and it is sent over downlink in 6TiSCH which is more constrained than uplink in 6TiSCH. Therefore only message_2 is discussed below.

    Connection IDs - It has been discussed to make conection IDs mandatory (or default). Given the current connection id encoding this would add 1 byte to message_2 when sent in CoAP response. Note that there has also been suggested to move connection ids to a shim layer. This would not affect the total overhead, just move bytes from EDHOC to lower layers. If connection ids are kept in EHOC I think EDHOC should adopt Christians suggestion for encoding.

    bstr_identifier - Is has been pointed out that the bstr_identifier is complicated and looks like overopimization, adding complexity to spec and implementations. Christian has some excellent suggestions on how to encode the connection identifiers. Then the bstr_identifier is only use to encode ID_CRED containing kid. This optimisations could maybe be skipped. This would add 1 byte to message_2 when certain 1 byte KIDs are used. I think this should be done unless the byte saving is very much needed.

[...]

PR #143, opened 2021-07-30
Optimization of message_2 to 45 bytes

emanjon:


emanjon:
message_2 could be made 44 bytes by using the known length of SIGNATURE_OR_MAC_2

marco-tiloca-sics:
I suggest to have C_R as the last element of the CBOR sequence (which shouldn't be a problem and still allows for a sequential sequence parsing), i.e.:

```
~~~~~~~~~~~ CDDL
data_2 = (
  G_Y_CIPHERTEXT_2 : bstr,
  C_R : bstr / int
)
~~~~~~~~~~~
```

When using CoAP (or other transports with the same correlation properties), this would preserve a current good pattern as useful for message recognition: if a response transporting an EDHOC message starts with a CBOR integer, then it's an EDHOC error message.

It would be good to preserve this, hence why keeping C_R at the end of the new message_2, since it can now be either a CBOR byte string or a CBOR integer.

gselander (approved):

gselander:
Merging this now as there was no objection to:
https://mailarchive.ietf.org/arch/msg/lake/wFREGTMBkR5LEcw3JAksJ782eCk/

This provides a solution to the concrete issue raised in #103.

gselander:
Mistake, I intended to merge, not close.

PR #176, opened 2021-09-22
Authentication parameters

emanjon:
Here is a rough first suggestion for changes.


gselander (approved):
Looks good. Two other things we discussed: 1. using credential "as received" instead of canonicalization. 2. Examples of multiple CBOR header maps.

Issue #189, opened 2021-10-29
Optional padding to hide length of ID_CRED_I and ID_CRED_R?


emanjon:
The confidentiality protection in EDHOC does not conceal the length of the identifiers ID_CRED_I and ID_CRED_R. We should maybe add OPTIONAL padding for deployments wanting to give eavesdroppers less information regarding the length of these identifiers. This could quite easily be achieved by allowign 0 or more simple values (e.g. true) in the beginning of the plaintexts.

```
OLD:
plaintext = ( ID_CRED_R / bstr / int, Signature_or_MAC_2, ? EAD_2 )

NEW:
plaintext = ( *true, ID_CRED_R / bstr / int, Signature_or_MAC_2, ? EAD_2 )

```

emanjon:
I made a PR. https://github.com/lake-wg/edhoc/pull/190

The document definitly need to information on that the length of identities can be leaked and that this can be mitigated by e.g., chosing fixed length identifier for all the Initiators.

I think OPTIONAL padding should be added for use cases like certificates. I am not sure it will be used very much, but if feels like a feature a security protocol is expected to have.

emanjon:
Based on Carsten's comment in COSE about failure from kid with int,

- My initial proposal was to make padding support for both sending and receiving optional.
- Marco commented on the PR and wanted mandatory support on the receiving side.
- Marco commented during the interim that he also wanted mandatory support on the sending side.

Before I update the PR, I think this should be discussed a bit more. There has earlier been comments from @chrysn and @StefanHri to not increase code size. I agree with Marco that mandatory to support for receiver makes sense. I am not sure about mandatory support for sender. My understanding that code for adding and removing padding would be different pieces of code.

marco-tiloca-sics:
I proposed:

- For the receiver, mandatory to support, since a padded message might legitimately come at any time.
- For the sender, optional to use, on a per-message basis. If the sender is surely never going to pad an outgoing message, no need for it to support padding as a sender at all, of course.

emanjon:
>For the sender, optional to use

I think the important question is if it is mandatory to implement.

gselander:
Currently the spec says
"Padding is OPTIONAL to support."

We proposed in the interim 

”OPTIONAL to support when sending 
MANDATORY to support when received"

What is the preferred text?

emanjon:
@marco-tiloca-sics 

The optional to use per message is already in the PR I think as "may be used".

[...]

PR #190, opened 2021-10-30
Padding and Privacy (issue #189)

emanjon:


draft-ietf-lake-edhoc.md
```diff
@@ -1090,9 +1094,11 @@ Implementations MAY support message_4. Error codes 1 and 2 MUST be supported.
 
 Implementations MAY support EAD.
 
+Padding is OPTIONAL to support.
```
marco-tiloca-sics, line 1097: Perhaps you mean "Including padding to an EDHOC message is OPTIONAL to support."

That is, I would expect the active use of padding from a sender node to be indeed optional.

However, a recipient node always has to be able to parse an EDHOC message, also in case padding is included. This is already what seems to be intended above in this PR, about the updated message format and processing.

emanjon: "That is, I would expect the active use of padding from a sender node to be indeed optional."
Yes, I think that is already in the PR.

"However, a recipient node always has to be able to parse an EDHOC message, also in case padding is included."
Not necerasy as discussed by Carsten during the last COSE meeting. You could live when the
recipient failing nicely. CBOR makes this possible. The recipient would know just now that does not expect
true in the beginning of the plaintext and fail. It does not need to know the meaning of the true. But
mandatory to support padding removal for recipient might be good.

gselander:
@emanjon 
@marco-tiloca-sics 

Can we merge this?

marco-tiloca-sics:
Yes, the latest updates look good.

PR #199, opened 2021-11-10
Updates following Marco's review

gselander:


emanjon (changes requested):
"By default, the message flow is as follows:"
- Text misses error as a result from 4, that is not very default, but none of the error are....


"but intended to enable the use short identifiers in combination with simplifying the retrieval of the right security context in the application protocol."

- of in "the use short" is missing
- Why would "enable" be better than "simplify"? Don't remember what EDHOC says, but simplify is absolutely a better word than enable for the use in COSE/OSCORE.
- Retrieval of the right security context is also for EDHOC. This change remove that by specifying "application protocol".


"identities to verify the 'sub' (subject) claim of the CWT."
- Should use the same text for X509 and CWT.
- Not added in this PR, but "sub" claim is just an example. CTW might add AltSub in the future.


"message_4 MUST be supported and the Responder MUST send message_4."
- Good change, but the "Responder MUST send message_4" is still strange, would about I? Maybe just "message_4 MUST be supported and used"?

gselander:
> 
> 
> "By default, the message flow is as follows:"
> 
>     Text misses error as a result from 4, that is not very default, but none of the error are....
>

Done

 
> "but intended to enable the use short identifiers in combination with simplifying the retrieval of the right security context in the application protocol."
> 
>     of in "the use short" is missing
>     Why would "enable" be better than "simplify"? Don't remember what EDHOC says, but simplify is absolutely a better word than enable for the use in COSE/OSCORE.
>     Retrieval of the right security context is also for EDHOC. This change remove that by specifying "application protocol".
>

Marco's comment was about interpretation of "security context", I reverted and bring up that in the response to his review mail.
 
>
> "identities to verify the 'sub' (subject) claim of the CWT."
> 
>     Should use the same text for X509 and CWT.
>     Not added in this PR, but "sub" claim is just an example. CTW might add AltSub in the future.
> 

Marco's comment was an editorial, which I kept. Your comment is a separate issue which i opened as #215

>
> "message_4 MUST be supported and the Responder MUST send message_4."
> 
>     Good change, but the "Responder MUST send message_4" is still strange, would about I? Maybe just "message_4 MUST be used"?
> 
> 

Done

PR #211, opened 2021-12-08
Updates following Stephen's review

gselander:


gselander:
Just first commit, more to come  . . .

emanjon (changes requested):
- Oxford comma
  OLD "X.509 and"
  NEW "X.509, and"

- I don't think this figure text is appropriate here. It mixes EDHOC and SIGMA, claim wrong things about SIGMA, and includes normative text in the caption. AEAD is defined earlier.
   OLD
   The SIGMA construction allows message_2 to include a plain encryption Enc(), whereas message_3 includes an authenticated encryption with additional data AEAD() that MUST NOT be replaced by a plain encryption. More details in Section 8.2."} {: artwork-align="center"}
   NEW
   The SIGMA construction allows message_2 to include a plain encryption Enc(), whereas message_3 is required to be authenticated encryption.

- I am not a big fan of the long caption in fig-error1 and fig-error2. For fig-error2 there is similar text "above" the figure "In the second example ({{fig-error2}})," ans in the caption....

gselander:
Merging this now to avoid conflicts when rewriting 3.5. Further comments are still welcome.

Issue #201, opened 2021-11-11
Minor cryptographic explanations


emanjon:
- The MAC length MUST be at least 8 bytes.
- Compact representation only for G_X and G_Y
- nonce for binding with the event that triggered KeyUpdate
- Explanation of no running hash

These have already seen commits to master

emanjon:
Add that less then 64 bit MACs shall not be used with EDHOC. The current text only refer to the MAC length parameter and not to COSE AEADs in general.

emanjon:
Add G_X collision text already in Responder message_1 processing. Based on Stephen's review this might be missed otherwise.

emanjon:
Security considerations on 64 and 128 bit MACs based on Stefans review

emanjon:
Some of the considerations in
https://mailarchive.ietf.org/arch/msg/lake/75nRaD6czYG6RqLT06Qe8C_lsaM/

Should likely be added.

emanjon:
The last parts was fixed in commit 678619c.

PR #220, opened 2021-12-21
Minor cryptographic explanations

emanjon:
#201


draft-ietf-lake-edhoc.md
```diff
@@ -1169,16 +1171,15 @@ An attacker can also send faked message_2, message_3, message_4, or error in an
...
+If two nodes unintentionally initiate two simultaneous EDHOC message exchanges with each other even if they only want to complete a single EDHOC message exchange, they MAY terminate the exchange with the lexicographically smallest G_X. If the two G_X values are equal, the received message_1 MUST be discarded to mitigate reflection attacks. Reflection attacks are not possible when the other endpoint's identity is verified, but can be a problem with trust on first use (TOFU) and could be a problem if methods with pre-shared keys are are specified in the future. Note that in the case of two simultaneous EDHOC exchanges where the nodes only complete one and where the nodes have different preferred cipher suites, an attacker can affect which of the two nodes’ preferred cipher suites will be used by blocking the other exchange.
```
pbtgit: "An attacker can affect which of the two nodes’ preferred cipher suites will be used by blocking the other exchange."
What are the exact pre-conditions for such an attack?

gselander: I opened a separate issue #226

emanjon: This is wrong, I don't know if it was true in an early version or if has always been wrong.

I tried to fix this in PR #231

emanjon:
This could not be merges for some reason so I made a separate commit 678619c. The commit also includes the fix to Paters comment above.

PR #225, opened 2022-01-10
Updates following review by Sean

gselander:


emanjon (changes requested):
>The algorithms of cipher suite 24 are included in the CNSA suite {{CNSA}}.

Sounds backwards. Maybe "Cipher suite 24 includes algorithms from the CNSA suite {{CNSA}}."


>transport of error messages,

Seems strange to single out error messages. The transport is responsible for transport of all types of EDHOC messages including error.....

gselander:
Merging this now to avoid conflicts when rewriting 3.5. Further comments are still welcome.

Issue #226, opened 2022-01-19
Blocking one of multiple simultaneous EDHOC exchanges


gselander:
Channeling question from @pbtgit on Section 8.8 (raised in #220 but worth a separate issue):

"Note that in the case of two simultaneous EDHOC exchanges where the nodes only complete one and where the nodes have different preferred cipher suites, an attacker can affect which of the two nodes’ preferred cipher suites will be used by blocking the other exchange."

What are the exact pre-conditions for such an attack?

emanjon:
This was fixed in commit 678619c.

PR #231, opened 2022-01-21
Blocking one of multiple simultaneous EDHOC exchanges (#226)

emanjon:
#226


emanjon:
This was fixed in commit 678619c.

Issue #214, opened 2021-12-10
Security considerations on gererating secret material and public material such as connection IDs.


emanjon:
Based on Stephens review

gselander:
Referring to this in #202:

> if some random values are visible 
> (connection identifiers?) then it can make sense to
> derive those from a different random stream compared to
> that used for randomly picking secrets. That way the
> publicly visible random numbers are less likely to leak
> information about the state of the PRNG used for secrets.

gselander:
From the PR #233:

"As explained by Krawczyk, any attack mitigated by the NIST requiment would mean that the KDF is fully broken and would have to be replaced anyway."

I didn't understand "any attack mitigated by the NIST requirement", and how mitigation could imply that something is broken.

As far as I understand, the statement is something like "if there is a problem deriving secret and non-secret material from the same KDF, then there is a problem with the KDF"

Perhaps "an attack based on the violation of the NIST requirement is an attack on the KDF itself"?

emanjon:
Yes, it was likely meants to say "if any attack is mitigate"

Let's reformulate according to your suggestion

emanjon:
I made a commit to correct spelling and grammar.

- I don't think we can give applications more guidance here. How to gererate e.g. private key in certificates are way outside of EDHOC, and it would anyway be hard to say something general as deployments might look very different.

- I did not want to say anything about randomly generated conenction ID as that is not how connection IDs is meant to be generated. They are intended to be as short as possible. An implentation is allowed to use long random bytestrings but is not recommended in general.

gselander:
PR merged

PR #233, opened 2022-01-22
Security considerations on gererating secret material and public material such as connection IDs. #214 

emanjon:
#214


emanjon:
I think this can be merged.

Issue #208, opened 2021-12-03
Error message => Discontinue


gselander:
From Marco's review:

[Section 5.2.3]

* "If an error message is sent, the session MUST be discontinued."

    Is this sentence actually required? Regardless sending an error 
message or not, the session is discontinued if any processing step 
fails, which is the reason for possibly sending an error message. Or are 
there situations where an error message is sent even if no processing 
step fails?

gselander:
This sentence does not talk about failed processing, it is a general implication stating the necessity of discontinuing the session if an error message is sent (for whatever reason), in contrast to the converse statement in the preceding sentence which states that there is no necessity of sending an error message if the session is discontinued.

But it is not explained why you would send an error unless processing has failed. Do we have any example?

emanjon:
I don't see the problem here at all. In general in security protocols there is no connection between some processing step failing, sending an alert, and terminationg the connection.

Error => processing fails is not true. It is not defined when an error is sent.

Close?

emanjon:
When to send an error is left for the implementation. Does that need to explained better?

emanjon:
Related comment from Sean

```
s5.2.3, s5.3.3, s5.4.3, s5.5.3, last sentence (question - probably being pedantic): If there is an error but an error message is not sent, is the session discontinued? How does the peer know it was discontinued if an error is not sent?
```

Let's go over and make sure the document answer

- When is the protocol discontinued?
- When is an error sent?
- How does the other party get information about discontinue and the type of error?

My understanding

- Up to the implementation/application when to discontinue the protocol. My feeling is that this is broader than "processing step fails"
- When the protocol is discontinued, an error MUST be sent, expect if that cannot be done due to DoS. The error tels the other party that the protocol has been discontinued and why it has been discontinued.
- A party receiving an error knows that the other party discontinues or that an attacker sent the error. In DoS situation the other party gets no information.

emanjon:
I will make a suggested PR for this in the upcoming weeks.

emanjon:
I made a PR to try to address the two comments from Marco and Sean which where quite different but both involve error.

As in HTTP and CoAP the intention is that error can be both "client errors" and "server errors". I removed duplicate text from the processing sections and expanded the text in the error section.

The PR likely needs some further update but is hopefully a good start for discussion.

emanjon:
EDHOC is not following the standard client error server error design of HTTP and CoAP. Is this intentional?

[...]

PR #234, opened 2022-01-23
 Error message => Discontinue #208

emanjon:
#208


emanjon:
I think this is ready to merge. This addresses Marcos and Seans comments. Remaining discussions should be handled by separate PRs.

Issue #241, opened 2022-01-31
Update text on identity protection


emanjon:
Current draft says:
"EDHOC protects the credential identifier of the Initiator against active attacks and the credential identifier of the Responder against passive attacks."

Should probably point out that active attack would be a new message_1 to the lower layer address as the other message_1.

PR #248, opened 2022-02-10
Identity protection attack description

emanjon:
#241

Good to describe that the known attack on Sigma is that the attacker create a new message_1, no attack is known where an active attacker modifies the ongoing connection (even if the lack of such attacks is likely not proven).


emanjon:
This seems ready to merge

Issue #253, opened 2022-02-22
Updates needed for traces-01


gselander:
I've started to work on the update of -traces, see c936328e.

What is missing is:

* Trace 2: SUITES_I = [6, 2] instead of 2.

* Trace 2: Add bytes for raw x.509 (in addition to the CBOR encoded, see c936328e)
     * In total two new entries: CRED_I (Raw Value) and CRED_R (Raw Value)
     
* Trace 2: Pretty print of all byte strings
   * the analogue of vectors.txt for p256
   * 16 bytes per row
    * space between bytes

* Marek's address

@stoprocent could you help out here?

There is also TODO on version and reference which we need to fix, but that's just an editorial.

gselander:
@stoprocent Submission deadline Monday in 10 days :-)

stoprocent:
@gselander I will work on this today in the afternoon.

stoprocent:
Hi

There was an incompatibility in the generator function for identifiers. This has been fixed here: https://github.com/stoprocent/edhoc/commit/abc197ef7372930c3c0d40dc4d0042cca27c6487
and updated test vectors are here: https://github.com/lake-wg/edhoc/pull/255

gselander:
Thanks, I'll take a look later today.

gselander:
New test vector included in trace 2.

gselander:
We are still missing update of label/remove of "Master" in trace 1, per #205.

@emanjon @stoprocent Could anyone of you help?

(Replace "OSCORE_Master_Secret/Salt" with "OSCORE_Secret/Salt" and re-generate test vector for EdDSA in vectors.txt)

marco-tiloca-sics:
> We are still missing update of label/remove of "Master" in trace 1, per https://github.com/lake-wg/edhoc/pull/205.
> 
> @emanjon @stoprocent Could anyone of you help?
> 
> (Replace "OSCORE_Master_Secret/Salt" with "OSCORE_Secret/Salt" and re-generate test vector for EdDSA in vectors.txt)

For the first trace using EdDSA, with the new labels "OSCORE_Secret" and "OSCORE_Salt", I derive:

OSCORE Master Secret (Raw Value) (16 bytes)
9565355973280280d64a3ce8d7fa0a93

OSCORE Master Salt (Raw Value) (8 bytes)
37478eddb09e5d7a

OSCORE Master Secret after KeyUpdate (Raw Value) (16 bytes)
[...]

PR #255, opened 2022-03-02
Fixing Connection Identifiers in p-256 vectors

stoprocent:


stoprocent:
This is addressing #253

PR #256, opened 2022-03-03
Updated test vectors to include G_X and G_Y compact representable keys.

stoprocent:
- Updated test vectors to include G_X and G_Y compact representable keys.
- Added uncompressed p256 G_X and G_Y public keys in vectors


stoprocent:
As usual updates to generator are here: https://github.com/stoprocent/edhoc/commit/92adddac2d808495c5c67b3058c84e98b551f247

malishav:
I summarize below my understanding of EC point representation and encoding.

In the affine representation, the point on an elliptic curve is defined as (X,Y), where X is the x coordinate, and Y is the y coordinate. There are 3 different ways to serialize a point:

## Compressed

`Output = Y || X`

`Y`: 1 byte that encodes the sign: `0x02` if y coordinate is positive; `0x03` if y coordinate is negative.
`X`: 32 bytes
Total serialized length: 33 bytes

Comment: Point `Y` is calculated from `X` based on the curve equation and the byte that encodes the sign. The procedure is at [1].

## Uncompressed

`Output = 0x04 || X || Y`

`0x04` - 1 byte
`X` - 32 bytes
`Y` - 32 bytes
Total serialized length: 65 bytes

Comment: Both points are fully encoded.

## Compact (called "raw" in the traces draft)

`Output = X`

`X`: 32 bytes
Total: 32 bytes

Comment: Defined in RFC6090, Section 4.2. Only the `X` coordinate is encoded. `Y` is calculated as either +Y or -Y.  Used in EDHOC. 
[1] https://www.secg.org/sec1-v2.pdf, Section 2.3.3

malishav:
The traces draft uses the term "raw" for what I believe is compact representation. I would propose that we only keep the compact representation in the draft, sicne this is what is used by EDHOC. 


@emanjon @gselander @franziskuskiefer : I tag you here to sanity check my understanding of the different representation and if you have any input related for this discussion

franziskuskiefer:
Thanks for tagging @malishav. The comment generally looks correct to me with one nit.

> Comment: Defined in RFC6090, Section 4.2. Only the `X` coordinate is encoded. `Y` is calculated as either +Y or -Y. Used in EDHOC.

`Y` is not either + or - as the edhoc draft specifies that the result is supposed to be equivalent to the compressed form with `0x02` for the sign byte (Appendix B. Compact Representation).

stoprocent:
@franziskuskiefer That's what I have said on the email thread with the guys. 

> Message 1 is a cbor sequence where G_X is placed as bstr not as COSE_key
> 0, [6, 2], h'0250A76B38EA840FA1B1A51152591D4CD52C75892152C870277225B1ED998ED953', -24
 
and the spec states that 

> For such use, the compact representation can be transformed into the SECG point compressed format by prepending it with the single byte 0x02 (i.e., M = 0x02 || X).

https://datatracker.ietf.org/doc/html/draft-ietf-lake-edhoc-12#appendix-B

malishav:
@stoprocent in my understanding, the text in the spec does not relate to the bytes on the wire, only explains how compact representation is transformed into uncompressed form for compliance with different libraries.

gselander:
@malishav is right, but I think I understand where the confusion may come from. 

Section 3.7 states:

>    EDHOC always uses compact representation of elliptic curve points,
>    see Appendix B. In COSE compact representation is achieved by
>    formatting the ECDH ephemeral public keys as COSE_Keys of type EC2 or
>    OKP according to Sections 7.1 and 7.2 of
>    [I-D.ietf-cose-rfc8152bis-algs], but only including the 'x' parameter
>    in G_X and G_Y. 

So this text does two things, it states that compact representation only includes the 'x' parameter, and it references appendix B. The former is what @malishav is referring to.

Now, appendix B first paragraph states

>    One example is ECDH with compact output, where only the x-coordinate
>    of the computed value is used as the shared secret.

and the rest of the appendix defines a format for the compact representation. 

The text @stoprocent quoted is prepended with a condition: 

>   If a y-coordinate is required for validation or compatibily with APIs the
>    value ~yp SHALL be set to zero.  For such use, the compact
>    representation can be transformed into the SECG point compressed
>    format by prepending it with the single byte 0x02 (i.e., M = 0x02 || X).

As stated in Section 3.7, the y-coordinate is not used, but appendix B may gives the impression that it should be used.

Is this the confusion?

stoprocent:
So to make it clear `G_X` and `G_Y` in vectors must be a raw `x` so `32b`?

gselander:
For the wire / message format, only use the 'x' parameter, i.e. compact format. For the test vectors I remember Marco and others have additionally requested both 'x' and 'y', i.e. the uncompressed format.

franziskuskiefer:
> For the wire / message format, only use the 'x' parameter, i.e. compact format. For the test vectors I remember Marco and others have additionally requested both 'x' and 'y', i.e. the uncompressed format.

👍🏻 Having the uncompressed point in addition to the compact form is nice because it allows testing the conversion. Not all P256 implementations will support compressed points such that EDHOC implementations will need to implement the conversion from compact representation to uncompressed.

But the test vectors must make clear whether the uncompressed point is the actual point or the point after converting from the compact representation since the sign gets lost.

stoprocent:
Please have a look. I've updated G_X and G_Y keys in vectors.

franziskuskiefer:
> Please have a look. I've updated G_X and G_Y keys in vectors.

Thanks. They look good. But I think a little more is needed. I presume `g_x_uncompressed_raw` is the actual point `g^x` (this must be noted somewhere).
Then we have a test vector here that covers the case where the sign bit happens to be "unchanged". At least one test vector where this is not the case is needed, i.e. where `g^x` is _not_ equal to `0x02 || X`

stoprocent:
`g_x_uncompressed_raw` is 0x04 || X || Y

marco-tiloca-sics:
Looks good! One comment:

> G_X (Raw Value) (Initiator's ephemeral public key, 'y'-coordinate) (32 bytes)

should probably just be

Initiator's ephemeral public key, 'y'-coordinate (32 bytes)

While this is part of the Initiator's ephemeral public key used for deriving the Diffie-Hellman secret (though not playing a role in that process), it is not part of the EDHOC G_X, which is about the 'x' coordinate only.

stoprocent:
Have a look in latest commit

gselander:
Looks good to me. 

Please all have a look at #258. Once we decided on that I think we are ready to update -traces.

franziskuskiefer:
> Have a look in latest commit

lgtm 👍🏻

marco-tiloca-sics:
I have just successfully validated the P-256 "Test Vector 1" from @stoprocent at https://github.com/stoprocent/edhoc/blob/test-vectors-rfc-p256/test-vectors-11/p256/vectors.txt

As a suggestion, it would be good to use the uncompressed notation also for the public authentication keys, specifying the 'x' and 'y' coordinates separately. That is:
```

G_R (Raw Value) (Responder's public authentication key, 'x'-coordinate) (32 bytes)
bb c3 49 60 52 6e a4 d3 2e 94 0c ad 2a 23 41 48 dd c2 17 91 a1 2a fb cb ac 93 62 20 46 dd 44 f0

G_R (Raw Value) (Responder's public authentication key, 'y'-coordinate) (32 bytes)
45 19 e2 57 23 6b 2a 0c e2 02 3f 09 31 f1 f3 86 ca 7a fd a6 4f cd e0 10 8c 22 4c 51 ea bf 60 72

G_I (Raw Value) (Initiator's public authentication key, 'x'-coordinate) (32 bytes)
8a 93 ca 7e 1b c8 46 47 d7 e7 eb 4c 61 07 c4 dc 4e 53 df 81 df d1 98 1c 7f 82 4a 7c 1b 61 a6 fc

G_I (Raw Value) (Initiator's public authentication key, 'y'-coordinate) (32 bytes)
91 36 28 13 c2 5d b6 af 93 be 22 c3 50 ce b2 51 89 5b 9f 3a 8d 85 a3 58 23 a2 22 2b 9d e2 c8 c8
```

gselander:
@marco-tiloca-sics Per your comment above I assume you mean the following headings for the 'y'-coordinates?

```
G_R (Raw Value) (Responder's public authentication key, 'x'-coordinate) (32 bytes)
bb c3 49 60 52 6e a4 d3 2e 94 0c ad 2a 23 41 48 dd c2 17 91 a1 2a fb cb ac 93 62 20 46 dd 44 f0

Responder's public authentication key, 'y'-coordinate (32 bytes)
45 19 e2 57 23 6b 2a 0c e2 02 3f 09 31 f1 f3 86 ca 7a fd a6 4f cd e0 10 8c 22 4c 51 ea bf 60 72

G_I (Raw Value) (Initiator's public authentication key, 'x'-coordinate) (32 bytes)
8a 93 ca 7e 1b c8 46 47 d7 e7 eb 4c 61 07 c4 dc 4e 53 df 81 df d1 98 1c 7f 82 4a 7c 1b 61 a6 fc

Initiator's public authentication key, 'y'-coordinate (32 bytes)
91 36 28 13 c2 5d b6 af 93 be 22 c3 50 ce b2 51 89 5b 9f 3a 8d 85 a3 58 23 a2 22 2b 9d e2 c8 c8
```

marco-tiloca-sics:
Right, that's consistent with the notation used for the ephemeral keys. I was too fast when copy-pasting :-)

stoprocent:
Replaced by #261

Issue #257, opened 2022-03-05
Clarification Appendix A.3.


StefanHri:
Hi,

I don't understand this very good: 
> "For the default case when the
   server is Responder, the prepended connection identifier is C_R, and
   C_I if the server is Initiator.  If message_1 is sent to the server,
   the CBOR simple value "true" (0xf5) is sent in its place (given that
   the server has not selected C_R yet)."

In the case where:
* the initiator is a client and
* the responder is a server
Do we need to replace C_I with 0xf5 in message_1??? 

@AleksanderDemianowski

gselander:
No, this is only referring to the *prepended* connection identifiers. 

The prepended connection identifiers are only used when the session can't be inferred from the previous message. The first message is a special case since there is no previous message, which is why "true" is used. 

Maybe the "in its place" is not sufficiently clear?

AleksanderDemianowski:
@StefanHri just for clarifiaction, @gselander please correct me if needed. Assuming the responder is a server, the first send data it's a message from client to server. This message is a concatenation of 1B (true - 0xf5) and EDHOC message_1. EDHOC message_1 includes C_I inside and this is not changing (is not repclaced). 
Always client (initiator) sends data like: (0xf5) || (METHOD, SUITES_I, G_X, C_I, EAD_1 ), where || is a symbol of concatenation.
This is what we can see as the first message on [Figure 10 in A.3. chapter.](https://tools.ietf.org/id/draft-ietf-lake-edhoc-12.html#fig-coap1)

gselander:
@AleksanderDemianowski Yes that is correct. 

Here is some new text replacing the existing description in an attempt to clarify further:

OLD
> In order to correlate a message received from a client to a message previously sent by the server, messages sent by the client are prepended with the CBOR serialization of the connection identifier which the server has chosen.  For the default case when the server is Responder, the prepended connection identifier is C_R, and C_I if the server is Initiator. If message_1 is sent to the server, the CBOR simple value "true" (0xf5) is sent in its place (given that the server has not selected C_R yet).


NEW
> In order to correlate a message received from a client to a message previously sent by the server in the same EDHOC session over CoAP, messages sent by the client are prepended with the CBOR serialization of the connection identifier which the server has chosen. This applies independently of if the CoAP server is Responder or Initiator. 

[...]

PR #259, opened 2022-03-16
Clarification of prepended identifiers, issue #257

gselander:


gselander:
This is being discussed at Hackathon now, so I'm merging to simplify for people to find the updated explanation.

Issue #167, opened 2021-09-10
Registration procedures for the new EDHOC registries


emanjon:
Currently a mixture of:
The registration procedure is "Expert Review".
The registration procedure is "Specification Required".

emanjon:
I changed EDHOC Error Code Registry to "Expert Review" to align with the others. It makes no sense that the Error registry has harder requirements than the methods or cipher suite registry,

We should maybe discuss Registration procedures for the new EDHOC registries at IETF 112.

gselander:
Reopen based on the comment by Kathleen

emanjon:
> At least new EDHOC method types would typically require specification.

Makes sense to me

emanjon:
Was is the status of this? This should be quick and easy to fix and close.

gselander:
The plan is to update Section 9.4 and replace "Expert Review" with "Specification Required". 

I noted now that in that section the columns of the registry are: 

> "Value, Description, and Reference" 

but the table with the referred initial contents (Figure 4) has columns:

> Method Type Value, Initiator Authentication Key, Responder Authentication Key, and Reference.

emanjon:
It should be 9.4

I think "Value, Description, and Reference" is correct, but just referencing Figure 4 does not work.

gselander:
We may need two tables. 
* Figure 4 doesn't need to have a Reference column. 
* We can make another table in 9.4 with "Value, Description, and Reference", where Description is "Intiator and Responder uses signature key for authentication" etc.

emanjon:
> Figure 4 doesn't need to have a Reference column.

A reference column is never needed. I would suggest to remove it from all the Tables in Section 9. That column is needed on the IANA homepage but does not help anybody in the EDHOC specification. IANA is smart enough to use the correct reference without us giving them a column.

emanjon:
Maybe Description should be separated into Initiator Authentication Key, Responder Authentication Key,

gselander:
The comments above about section 9.4 refers to 9.3 [EDHOC Method Type Registry] in the current master branch.

gselander:
An attempt to address all comments in PR #260.

I kept the column for References for the registries where a specification is required, method (9.3) and EAD (9.5).

gselander:
References column removed also in cases where spec. is required in #260, which is now merged.

PR #260, opened 2022-03-16
Update to IANA registration, issue #167

gselander:


emanjon:
I am against adding a stupid reference column.

Issue #262, opened 2022-03-20
Usage of Content-Format in CoAP when transferring EDHOC over CoAP


dnav:
As specified in section [A.3 Transferring EDHOC over CoAP](https://lake-wg.github.io/edhoc/draft-ietf-lake-edhoc.html#name-transferring-edhoc-over-coa), the payload in some CoAP messages can be an EDHOC message prepended by either a C_x identifier or "true". The draft states:

> The application/edhoc media type does not apply to these messages; their media type is unnamed.

However, the payload is actually a CBOR Sequence as defined by [RFC 8742](https://datatracker.ietf.org/doc/html/rfc8742). Thus, we could define and use a new media type "application/edhoc+cbor-seq".

Note that the usage of the Content-Format option is optional.

gselander:
Makes sense to me. 

@cabo, @chrysn, others: Comments?

Good that you bring up the optionality of Content-Format, we should stress that in this text.

The discussion in Hackathon also reminded me about the common-option-compression which @chrysn presented in the May 12 interim 2021 - how to replace the Uri-Path: "/.well-known/edhoc" with a, say,  integer valued CoAP option.

cabo:
Makes a lot of sense to me, too.

PR #263, opened 2022-03-20
Define new media type application/edhoc+cbor-seq, issue #262

dnav:
Signed-off-by: David Navarro <david.navarro@ioterop.com>

### DIFF
--- a/draft-ietf-lake-edhoc.md
+++ b/draft-ietf-lake-edhoc.md
@@ -303,7 +303,7 @@ To simplify for implementors, the use of CBOR and COSE in EDHOC is summarized in
 
 The EDHOC protocol consists of three mandatory messages (message_1, message_2, message_3) between Initiator and Responder, an optional fourth message (message_4), and an error message. All EDHOC messages are CBOR Sequences {{RFC8742}}. {{fig-flow}} illustrates an EDHOC message flow with the optional fourth message as well as the content of each message. The protocol elements in the figure are introduced in {{overview}} and {{asym}}. Message formatting and processing are specified in {{asym}} and {{error}}.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-lake-edhoc.md
```diff
@@ -1499,17 +1501,59 @@ IANA has added the media type "application/edhoc" to the "Media Types" registry.
...
+## CoAP Content-Formats Registry {#content-format}
+
+IANA has added the media types "application/edhoc" and "application/edhoc+cbor-seq" to the "CoAP Content-Formats" registry under the group name "Constrained RESTful Environments (CoRE) Parameters".
+
+~~~~~~~~~~~
++----------------------------+----------+-------+------------------------+
+| Media Type                 | Encoding | ID    | Reference              |
++----------------------------+----------+-------+------------------------+
+| application/edhoc          | -        | TBD42 | \[\[this document\]\]  |
+| application/edhoc+cbor-seq | -        | TBD43 | \[\[this document\]\]  |
```
emanjon, line 1553: Not my area of expertise. So I don't have any strong opinions.

The need seem to have a different media type for "connection ID", EDHOC

But what is the difference between "application/edhoc" and "application/edhoc+cbor-seq"? Do we need both?

- EDHOC (application/edhoc) is already always cbor-seq. 
- My understanding is that is that "edhoc+cbor-seq" means that this is edhoc but cbor-seq encoded.

My understanding would that that it would make more sense with

edhoc+cbor-seq
cid-edhoc+cbor-seq

(Feel free to ignore this comments)

gselander:
@cabo @chrysn and others: Any response to John's comment?

draft-ietf-lake-edhoc.md
dnav: Thank you for your comment. I went for the easiest path but your suggestion totally makes sense.

I updated the PR accordingly.

dnav:
@gselander I included John's suggestion in the PR.

gselander:
Looks good to me. Other opinions?

gselander (approved):
I find this distinction good and the solution consequential, still open for comments.

Issue #265, opened 2022-03-31
kid2 => int, instead of kid => bstr / int  


gselander:
There has been for some time a discussion in COSE about how to define integer-valued key identifiers:

https://mailarchive.ietf.org/arch/msg/cose/K0eifKhCluNreVwnJv2-LmhfLo8/

The extension of kid to bstr / int was previously agreed on the COSE mailing list but that is now revisited because of potential impact on existing COSE implementations. Three alternatives considered: 

Solution A (currently assumed by EDHOC).
kid => int / bytes

Solution B.
kid => bytes
kid2 => int / bytes

Solution C.
kid => bytes
kid2 => int

At the time of writing there is one voice in preference for solution C and no objections.

Does anyone see any issues with solutions B or C (besides minor change in current implementations)?

gselander:
@cabo made a proposal in the COSE mail thread above going into another direction [1]:

> all kids are byte strings and the integer representation is just a compressed way to express such a byte string

The idea is that certain byte strings (48 to be precise) should have integer representations in the interval -24, ... ,23 (which all have 1-byte CBOR encoding: 37, 36, ..., 21, 20, 00, 01, ..., 17)

In his mail [1] @cabo made a proposal for which these byte strings should be.
 
[1] https://mailarchive.ietf.org/arch/msg/cose/1EXuPcjrGWSXNTk4rw2Ffp6Jpnk/

After an offlist discussion that part of the proposal was changed so that the byte strings which have integer representation are those for which there is an integer in the interval -24..23 that happens to have this byte string as its CBOR encoding, i.e.:

h'37', h'36', ..., h'21', h'20', h’00’, h'01', ..., h’17'

gselander:
Following this line of thought, if C_x are ever only byte strings, then the mapping of Connection Identifiers to OSCORE Sender/Recipient ID can be simplified; the Sender/Recipient ID equals the byte string. E.g. if C_R is h'0a' in diagnostic notation then Sender ID = 0x0a.

The only complication is the format of EDHOC on the wire: 

```
C_x                 EDHOC           
bstr             on the wire           

h’’                    40                 
h’00’                  00                    
. . .
h’09’                  09                    
h’0a’                  10                      
. . .
h’17’                  23                       
h’18’               41 18                     
h’19’               41 19                      
h’20’                  -1                   
. . .
h’37’                 -24                   
h’38’               41 38                     
. . .
h’4711’          42 47 11

gselander:
[...]

PR #266, opened 2022-03-31
1-byte representation of bstr

gselander:


gselander:
Refers to #265. First draft.

Only last two commits are relevant here, the rest was left-overs from debugging. I may make a clean version later, but I don't see any trace of those changes in "Files changed".

gselander:
Replaced by #267, same changes on -edhoc but that PR does not have the traces-related commits that accidently got included in this one.

PR #267, opened 2022-04-01
Representation of bstr as 1-byte int

gselander:
First draft to address #265

### DIFF
--- a/draft-ietf-lake-edhoc.md
+++ b/draft-ietf-lake-edhoc.md
@@ -474,7 +474,7 @@ An example of a CRED_x is shown below:
   8 : {                                        /cnf/
     1 : {                                      /COSE_Key/
       1 : 1,                                   /kty/
-      2 : 0,                                   /kid/
+      2 : h'00',                               /kid/
      -1 : 4,                                   /crv/
      -2 : h'b1a3e89460e88d3a8d54211dc95f0b90   /x/
             3ff205eb71912d6db8f4af980d2db83a'
@@ -482,7 +482,7 @@ An example of a CRED_x is shown below:
   }
 }
 ~~~~~~~~~~~
-{: title="CWT Claims Set (CCS) containing an X25519 static Diffie-Hellman key and an EUI-64 identity. Integer representation of kid is assumed."}
+{: title="CWT Claims Set (CCS) containing an X25519 static Diffie-Hellman key and an EUI-64 identity."}
 
 
 ### Identification of Credentials {#id_cred}
--- a/draft-ietf-lake-edhoc.md
+++ b/draft-ietf-lake-edhoc.md
@@ -499,7 +499,7 @@ Example: CWT or CCS can be identified by a key identifier using the 'kid' parame
 
 * ID_CRED_x = { 4 : key_id_x }, where key_id_x : kid, for x = I or R.
 
-The COSE 'kid' parameter has byte string as value. To allow more one-byte key identifiers, which may be useful in many scenarios since constrained devices only have a few keys, we use the same integer representation as for connection identifiers, see {{bstr-repr}}. As stated in Section 3.1 of {{I-D.ietf-cose-rfc8152bis-struct}}, applications MUST NOT assume that 'kid' values are unique and several keys associated with a 'kid' may need to be checked before the correct one is found. Applications might use additional information such as 'kid context' or lower layers to determine which key to try first. Applications should strive to make ID_CRED_x as unique as possible, since the recipient may otherwise have to try several keys.
+The value of a COSE 'kid' parameter is a byte string. To allow more one-byte key identifiers, which may be useful in many scenarios since constrained devices only have a few keys, we use the same integer representation as for connection identifiers, see {{bstr-repr}}. As stated in Section 3.1 of {{I-D.ietf-cose-rfc8152bis-struct}}, applications MUST NOT assume that 'kid' values are unique and several keys associated with a 'kid' may need to be checked before the correct one is found. Applications might use additional information such as 'kid context' or lower layers to determine which key to try first. Applications should strive to make ID_CRED_x as unique as possible, since the recipient may otherwise have to try several keys.
 
 See {{COSE}} for more examples.
 
--- a/draft-ietf-lake-edhoc.md
+++ b/draft-ietf-lake-edhoc.md
@@ -375,7 +375,9 @@ For example, a byte string valued C_I equal to 0xFF (0x41FF in CBOR encoding) is
 
 ### Representation of Byte String Identifiers {#bstr-repr}
 
-The integers -24, ..., 23 are all CBOR encoded as one byte, see {{fig-int-one-byte}}.
+To allow byte string identifiers with minimal overhead on the wire, certain byte strings are defined to have integer representations. These byte strings are those for which there is an integer that happens to have this byte string as its CBOR encoding.
+
+Note that the integers -24, ..., 23 are all CBOR encoded as one byte, see {{fig-int-one-byte}}.
 
 ~~~~~~~~~~~
 Integer:               -24   -23  ...   -2    -1     0     1  ...   23
@@ -384,14 +386,21 @@ CBOR encoding (1 byte): 37    36  ...   21    20    00    01  ...   17
 {: #fig-int-one-byte title="One-Byte CBOR Encoded Integers"}
 {: artwork-align="center"}
 
-To allow more byte string identifiers that are only one byte on the wire, certain byte strings are defined to  have integer representations in the interval -24, ... ,23. These byte strings are those for which there is an integer in the interval -24, ..., 23 that happens to have this byte string as its CBOR encoding, see {{fig-bstr-as-int}} and compare {{fig-int-one-byte}}.
+The byte strings which coincide with the CBOR encoding of an integer are thus not represented by their CBOR byte string encoding, but with the CBOR int encoding of the corresponding integer.
+
+For example, h'21' is represented by 0x21 (CBOR encoding of the integer -2), not by 0x4121;  h'189B' is represented by 0x189B (CBOR encoding of the integer 155) not by 0x42189B. The byte strings which have one byte integer representations are shown in {{fig-bstr-as-int}}.
 
 ~~~~~~~~~~~
-Byte string           h'37' h'36' ... h'21' h'20' h'00' h'01' ... h'17'
+Byte string:          h'37' h'36' ... h'21' h'20' h'00' h'01' ... h'17'
 Integer:               -24   -23  ...   -2    -1     0     1  ...   23
+CBOR encoding:          37    36  ...   21    20    00    01  ...   17
 ~~~~~~~~~~~
 {: #fig-bstr-as-int title="Byte Strings Represented as Integers"}
 {: artwork-align="center"}
+
+Byte strings which do not correspond to CBOR encoding of integers are encoded as normal CBOR byte strings. For example, h'A5' is represented by 0x41A5; h'4711' is represented by 0x424711.
+
+One way to view this representation of byte strings is as a transport encoding. A byte string which parses as a CBOR int is just copied directly into the message, a byte string which doesn't parse as a CBOR int is encoded as a CBOR bstr during transport.
 
 
 ## Transport {#transport}
@@ -472,7 +481,7 @@ An example of a CRED_x is shown below:
   }
 }
 ~~~~~~~~~~~
-{: title="CWT Claims Set (CCS) containing an X25519 static Diffie-Hellman key and an EUI-64 identity."}
+{: title="CWT Claims Set (CCS) containing an X25519 static Diffie-Hellman key and an EUI-64 identity. Integer representation of kid is assumed."}
 
 
 ### Identification of Credentials {#id_cred}
@@ -775,7 +784,7 @@ message_1 = (
   METHOD : int,
   SUITES_I : suites,
   G_X : bstr,
-  C_I : bstr,
+  C_I : bstr / int,
   ? EAD_1 : ead,
 )
 
@@ -787,7 +796,7 @@ where:
 * METHOD - authentication method, see {{method}}.
 * SUITES_I - array of cipher suites which the Initiator supports in order of preference, the first cipher suite in network byte order is the most preferred by I, the last is the one selected by I for this session. If the most preferred cipher suite is selected then SUITES_I contains only that cipher suite and is encoded as an int. The processing steps are detailed below and in {{wrong-selected}}.
 * G_X - the ephemeral public key of the Initiator
-* C_I - variable length connection identifier
+* C_I - variable length connection identifier. Note that connection identifiers are byte strings but may be represented as integers in the message, see {{bstr-repr}}.
 * EAD_1 - external authorization data, see {{AD}}.
 
 ### Initiator Processing of Message 1 {#init-proc-msg1}
@@ -828,14 +837,14 @@ message_2 SHALL be a CBOR Sequence (see {{CBOR}}) as defined below
 ~~~~~~~~~~~ CDDL
 message_2 = (
   G_Y_CIPHERTEXT_2 : bstr,
-  C_R : bstr,
+  C_R : bstr / int,
 )
 ~~~~~~~~~~~
 
 where:
 
-* G_Y_CIPHERTEXT_2 - the concatenation of G_Y (i.e., the ephemeral public key of the Responder) and CIPHERTEXT_2
-* C_R - variable length connection identifier
+* G_Y_CIPHERTEXT_2 - the concatenation of G_Y (i.e., the ephemeral public key of the Responder) and CIPHERTEXT_2.
+* C_R - variable length connection identifier. Note that connection identifiers are byte strings but may be represented as integers in the message, see {{bstr-repr}}.
 
 ### Responder Processing of Message 2 {#asym-msg2-proc}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-lake-edhoc.md
```diff
@@ -358,7 +358,8 @@ EDHOC includes the selection of connection identifiers (C_I, C_R) identifying a
 
 Connection identifiers may be used to correlate EDHOC messages and facilitate the retrieval of protocol state during EDHOC execution (see {{transport}}) or in subsequent applications of  EDHOC, e.g., in OSCORE (see {{ci-oscore}}). The connection identifiers do not have any cryptographic purpose in EDHOC.
 
-Connection identifiers in EDHOC are byte strings or integers, encoded in CBOR. One byte connection identifiers (the integers -24 to 23 and the empty CBOR byte string h'') are realistic in many scenarios as most constrained devices only have a few connections. An endpoint may choose to select only integer or only byte string connection identifiers.
+Connection identifiers in EDHOC are CBOR byte strings. Since most constrained devices only have a few connections short identifiers are desirable in many cases. However, except for the empty byte string h'' which encodes as as one byte (0x40), all byte strings are CBOR encoded as two or more bytes. Therefore EDHOC specifies certain byte strings to be represented as CBOR ints on the wire, see {{bstr-repr}}.
```
cabo, line 361:
```suggestion
Connection identifiers in EDHOC are CBOR byte strings. Since most constrained devices only have a few connections, short identifiers are desirable in many cases. However, except for the empty byte string h'' which encodes as as one byte (0x40), all byte strings are CBOR encoded as two or more bytes. Therefore EDHOC specifies certain byte strings to be represented as CBOR ints on the wire, see {{bstr-repr}}.
``` 

(add comma for readability)

draft-ietf-lake-edhoc.md
```diff
@@ -368,7 +369,29 @@ If connection identifiers are used by an application protocol for which EDHOC es
 
 ### Use of Connection Identifiers with OSCORE {#ci-oscore}
 
-For OSCORE, the choice of connection identifier results in the endpoint selecting its Recipient ID, see Section 3.1 of {{RFC8613}}, for which certain uniqueness requirements apply, see Section 3.3 of {{RFC8613}}. Therefore, the Initiator and the Responder MUST NOT select connection identifiers such that it results in same OSCORE Recipient ID. Since the Recipient ID is a byte string and an EDHOC connection identifier is either a CBOR byte string or a CBOR integer, care must be taken when selecting the connection identifiers and converting them to Recipient IDs. A mapping from EDHOC connection identifier to OSCORE Recipient ID is specified in {{edhoc-to-oscore}}.
+For OSCORE, the choice of connection identifier results in the endpoint selecting its Recipient ID, see Section 3.1 of {{RFC8613}}, for which certain uniqueness requirements apply, see Section 3.3 of {{RFC8613}}. Therefore, the Initiator and the Responder MUST NOT select connection identifiers such that it results in same OSCORE Recipient ID. Since the connection identifier is byte-valued, it is converted to an OSCORE Recipient ID equal to the byte string.
```
cabo, line 372: "byte-valued"?  I thought it was a byte string.

draft-ietf-lake-edhoc.md
```diff
@@ -476,7 +499,7 @@ Example: CWT or CCS can be identified by a key identifier using the 'kid' parame
...
+The COSE 'kid' parameter has byte string as value. To allow more one-byte key identifiers, which may be useful in many scenarios since constrained devices only have a few keys, we use the same integer representation as for connection identifiers, see {{bstr-repr}}. As stated in Section 3.1 of {{I-D.ietf-cose-rfc8152bis-struct}}, applications MUST NOT assume that 'kid' values are unique and several keys associated with a 'kid' may need to be checked before the correct one is found. Applications might use additional information such as 'kid context' or lower layers to determine which key to try first. Applications should strive to make ID_CRED_x as unique as possible, since the recipient may otherwise have to try several keys.
```
cabo, line 502:
```suggestion
The value of a COSE 'kid' parameter is a byte string. To allow more one-byte key identifiers, which may be useful in many scenarios since constrained devices only have a few keys, we use the same integer representation as for connection identifiers, see {{bstr-repr}}. As stated in Section 3.1 of {{I-D.ietf-cose-rfc8152bis-struct}}, applications MUST NOT assume that 'kid' values are unique and several keys associated with a 'kid' may need to be checked before the correct one is found. Applications might use additional information such as 'kid context' or lower layers to determine which key to try first. Applications should strive to make ID_CRED_x as unique as possible, since the recipient may otherwise have to try several keys.
```

draft-ietf-lake-edhoc.md
```diff
@@ -752,7 +775,7 @@ message_1 = (
   METHOD : int,
   SUITES_I : suites,
   G_X : bstr,
-  C_I : bstr / int,
+  C_I : bstr,
```
cabo, line 778: Why this change?

draft-ietf-lake-edhoc.md
```diff
@@ -805,7 +828,7 @@ message_2 SHALL be a CBOR Sequence (see {{CBOR}}) as defined below
 ~~~~~~~~~~~ CDDL
 message_2 = (
   G_Y_CIPHERTEXT_2 : bstr,
-  C_R : bstr / int,
+  C_R : bstr,
```
cabo, line 831: Why this change?

cabo (commented):
Here's a quick leaf-through.

gselander:
@cabo Thanks. I actually thought about the text parts you commented on but I had worked too late to do something about it.

About the connection id type It should be 

C_x : bstr / -24..23,

or do you see any reason to keep bstr / int?

cabo:
Maybe I would give `-24..23` a name... Not enough creativity today.  `obint` (one-byte int)?  Haven't checked that in all languages :-)
Yes, that narrow range is a more accurate representation for what you are saying in the text.

draft-ietf-lake-edhoc.md
gselander: Changed in 0ab48af

gselander:
Right, let's first hear what people think.

draft-ietf-lake-edhoc.md
marco-tiloca-sics, line 778: Also consistently with the later editing in the "CDDL Definitions" appendix, this should be

`C_I : bstr / -24..23,`

draft-ietf-lake-edhoc.md
marco-tiloca-sics, line 831: Also consistently with the later editing in the "CDDL Definitions" appendix, this should be

`C_R : bstr / -24..23,`

draft-ietf-lake-edhoc.md
```diff
@@ -368,7 +369,29 @@ If connection identifiers are used by an application protocol for which EDHOC es
...
+
+~~~~~~~~~~~
+Integer:               -24   -23  ...   -2    -1     0     1  ...   23
+CBOR encoding (1 byte): 37    36  ...   21    20    00    01  ...   17
+~~~~~~~~~~~
+{: #fig-int-one-byte title="One-Byte CBOR Encoded Integers"}
+{: artwork-align="center"}
+
+To allow more byte string identifiers that are only one byte on the wire, certain byte strings are defined to  have integer representations in the interval -24, ... ,23. These byte strings are those for which there is an integer in the interval -24, ..., 23 that happens to have this byte string as its CBOR encoding, see {{fig-bstr-as-int}} and compare {{fig-int-one-byte}}.
+
```
marco-tiloca-sics, line 388: Is there a reason to limit this representation only to the interval -24..23?

If it rather considered all the integer values:

* There will be several more identifiers that can be transported as integers, again with 1 byte less overhead compared to the byte string representation. For example, the identifier 0x1839 can be transported as 0x1839 (CBOR integer 57) rather than 0x421839 (CBOR byte string).

* I believe it would be overall simpler for an implementation, that would simply have to check about integer vs. byte string, rather than specifically integer -24..23 vs. byte string.

The same applies for 'kid' values when 'kid' is used in ID_CRED_X.

gselander: Without commenting on this proposal, I don't think we should optimise the sheer number of short identifiers.
There needs to be a balance here between what is easily specified, understood and coded. 

IMO, as long as we can have a number of 1-byte connection identifiers and key identifiers in EDHOC (to comply with the LAKE requirements) and support for the empty byte string in OSCORE (to reduce bytes in the most frequent message exchange) the main effort should be spent on making the EDHOC specification and the conversion between EDHOC and OSCORE identifiers as simple as possible. 

@marco-tiloca-sics Have you written down the procedure for reversing the representation, and for mapping back and forth between EDHOC and OSCORE?

One concern I have (with #267 in general) is with the pre-pending of connection identifiers done at transport layer. If we introduce special EDHOC representations, does the transport layer also need to know about that? (This was one reason for doing #274.)

draft-ietf-lake-edhoc.md
gselander: Overtaken by later proposals

draft-ietf-lake-edhoc.md
gselander: Overtaken by later proposals

draft-ietf-lake-edhoc.md
gselander: The proposal to apply the mapping for all integers is now included. I agree that the user of transport needs to provide the transport-ready representation, including prepended identifier, this is also noted in the update.

gselander:
I think the PR is ready for review (thanks @cabo and @marco-tiloca-sics).

Short summary: we may not be able to register `kid` as bstr / int. This proposal is based on `kid` (and connection identifiers) intrinsically being byte strings, but can be represented as CBOR ints in EDHOC messages.

@chrysn @malishav @TimothyClaeys and others, please have a look.

cabo:
I don't think enabling the shorter integer representation for all byte strings that parse as CBOR integers is worth it.

By staying with -24..23, the integerness can easily be tested, as in

b.size == 1 && !(b[0]&0xC0) && ((b[0] & 0x1f) < 0x18)

and the integer can be extracted as

(b[0] & 0x1f) ^ -((b[0] >> 4) & 0x1)

(untested code, but probably close).

Allowing all integers means that the CBOR library needs to be called and there needs to be a way in that to handle unsuccessful results.

All for a quite questionable additional benefit.

cabo:
(Assuming the conversion to integer is mandatory -- if it is not, this becomes an interop testing nightmare.)

gselander:
Thanks @cabo. The changes to any length of integer representation was made for perceived simplicity by @marco-tiloca-sics. And yes, we must clarify that this is mandatory.

chrysn:
Let's get some numbers on what we get from just-bytes (0), -24..23 (1) vs. any-integer (2):

```
+---------+-+---------+-----------+-----------+
| size    | usable 0  | usable 1  | usable 2  |
+---------+-+---------+-----------+-----------+
| 1       | 1         | 49        | 49        |
| 2       | 256       | 208       | 720       |
| 3       | 64k       | 64k - 48  | ~3 x 64k  |
+---------+-+---------+-----------+-----------+
```

Clearly using any-integer is higher density (and I admit to having been the one who started going about these densities initially), but really where is it relevant? A system needs to use more than 257 identifiers before 1 vs 2 starts making a difference. And even then, things start to differ only for the EDHOC phase, there are (as there have always been) 256 1-byte and 64k 2-byte IDs for available for OSCORE, it's just that some of them (most of them, in either case) need an extra byte to inhabit CBOR encodings.

Constrained systems likely get away with picking their own identifiers exclusively from -24..23 anyway.

I think the larger question is "what is simpler?", and that depends by the use case:

* Picking a value from a series of pools (sorted by preference for OSCORE use and then preference for EDHOC use) unless it is taken:
  1. pools are h'', (-24..=23), h'00' to h'ff' with two holes, and the long strings.
  2. pools are h'', (-24..=23), h'00' to h'ff' with two holes, (-256..=-25), (24..256), h'0000' to h'ffff' with four holes etc.
* Encoding a picked value is trivial either way.
* Verifying that the peer encoded its ID correctly -- does the party receiving the C_x really need to do this?
* Decoding a peer's byte encoded C_x -- again, does this even ever need to be done? (The OSCORE contexts are best stored indexed by recipient ID)

At least if the latest two are answered "no" (but probably either way), I'd say that 1 is easier to implement, and while the coding density differs by a factor of ~3 for the 2-3 byte values, it's the same where it matters the most.

draft-ietf-lake-edhoc.md
gselander: Now reverted to one-byte integers only.

gselander:
Sorry for the delay, now reverted to (1).

I didn't try to make the CDDL exclude the disallowed byte strings, but made a normative statement. 

@cabo @marco-tiloca-sics @chrysn Please check.

marco-tiloca-sics:
It looks good to me.

malishav:
The PR looks quite clear to me.

draft-ietf-lake-edhoc.md
```diff
@@ -358,17 +358,46 @@ EDHOC includes the selection of connection identifiers (C_I, C_R) identifying a
...
 
 C_I and C_R are chosen by I and R, respectively. The Initiator selects C_I and sends it in message_1 for the Responder to use as a reference to the connection in communications with the Initiator. The Responder selects C_R and sends it in message_2 for the Initiator to use as a reference to the connection in communications with the Responder.
 
 If connection identifiers are used by an application protocol for which EDHOC establishes keys then the selected connection identifiers SHALL adhere to the requirements for that protocol, see {{ci-oscore}} for an example.
 
+### Representation of Byte String Identifiers {#bstr-repr}
+
+To allow identifiers with minimal overhead on the wire, certain byte strings are defined to have integer representations.
+
+The integers with one-byte CBOR encoding are -24, ..., 23, see {{fig-int-one-byte}}. This correspondence between integers and byte strings is a natural mapping between the byte strings with CBOR diagnostic notation h'00', h'01', ..., h'37' (except h'18' and h'19') and integers which are CBOR encoded as one byte.
```
emanjon, line 374: > is a natural mapping between the byte strings with CBOR diagnostic notation h'00', h'01', ..., h'37' (except h'18' and h'19') and integers which are CBOR encoded as one byte.

There are 54 such byte strings and 48 integers which are CBOR encoded as one byte

draft-ietf-lake-edhoc.md
```diff
@@ -476,7 +505,9 @@ Example: CWT or CCS can be identified by a key identifier using the 'kid' parame
 
 * ID_CRED_x = { 4 : key_id_x }, where key_id_x : kid, for x = I or R.
 
-Note that 'kid' is extended to support int values to allow more one-byte identifiers (see {{kid-header-param}} and {{kid-key-common-param}}) which may be useful in many scenarios since constrained devices only have a few keys. As stated in Section 3.1 of {{I-D.ietf-cose-rfc8152bis-struct}}, applications MUST NOT assume that 'kid' values are unique and several keys associated with a 'kid' may need to be checked before the correct one is found. Applications might use additional information such as 'kid context' or lower layers to determine which key to try first. Applications should strive to make ID_CRED_x as unique as possible, since the recipient may otherwise have to try several keys.
+The value of a COSE 'kid' parameter is a byte string. To allow one-byte encodings of ID_CRED_x with key identifiers 'kid', which is useful in scenarios with only a few keys, the integer representation of identifiers in {{bstr-repr}} is applied. For details, see {{asym-msg2-proc}} and {{asym-msg3-proc}}.
```
emanjon, line 508: There should be a MUST here.

draft-ietf-lake-edhoc.md
```diff
@@ -1122,7 +1153,9 @@ An implementation MAY support only Initiator or only Responder.
 
 An implementation MAY support only a single method. None of the methods are mandatory-to-implement.
 
-Implementations MUST support 'kid' parameters of type int. None of the other COSE header parameters are mandatory-to-implement.
+Implementations MUST support byte string identifiers represented as one-byte CBOR ints, as described in {{bstr-repr}}.
```
emanjon, line 1156: > Implementations MUST support byte string identifiers represented as one-byte CBOR ints, as described in {{bstr-repr}}.

If we have MUST earlier in the document I think this sentence should be removed. "supported" gives the idea that this is not mandatory to encode as int, which would lead to a non-deterministic encoding.

draft-ietf-lake-edhoc.md
```diff
@@ -1202,7 +1235,7 @@ An attacker observing network traffic may use connection identifiers sent in cle
...
 
-Information regarding the lengths of ID_CRED_I, ID_CRED_R, EAD_2, EAD_3, and EAD_4 are leaked to an passive attacker. Many COSE header parameters have variable length. To mitigate an attacker from differentiating endpoints with identifiers of different length, and implementation may e.g., only use fix length identifiers like 'kid' of length 1 for the Responders. Alternatively padding may be used to hide the length of e.g., certificates by value in 'x5chain' or 'c5c'.
+Information regarding the lengths of ID_CRED_I, ID_CRED_R, EAD_2, EAD_3, and EAD_4 are leaked to a passive attacker. Many COSE header parameters have variable length. To mitigate an attacker from differentiating endpoints with identifiers of different length, an implementation may, e.g., only use fix length identifiers like 'kid' with one-byte CBOR int representations for the Responders. Alternatively padding may be used to hide the length of, e.g., certificates by value in 'x5chain' or 'c5c'.
```
emanjon, line 1238: > are leaked to a passive attacker.

are is too strong here

emanjon (approved):

draft-ietf-lake-edhoc.md
gselander: Addressed by #282

PR #274, opened 2022-04-02
First draft of separate integer key identifier

gselander:
This PR is an alternative to #267 for resolving #265. A separate integer key identifier called 'intkid' is used.


gselander:
Closing this in favor of the proposed change to #267.

Issue #270, opened 2022-04-01
Strengthen the TEE implementation


charlie-j:
In method 0, the attacker must be able to compromise the long term authentication key to break the authentication of the protocol. In other terms, leaking the ephemeral share and the session key is not enough to break authentication, and does only break secrecy. This leads to a situation where it makes sense to protect the long term authentication key inside a TEE, as it strengthens authentication.

The issue is that in other methods, where an agent authenticates with a long term DH key, the attacker only needs access to ephemeral and session keys to impersonate an agent. In turn, this implies that using a TEE for the long term DH share does not currently have any meaningful impact on the security of the protocol.

In more detail, for instance in Method 2, the MAC key used to authenticate the initiator is actually the final session key. An attacker can thus by learning the ephemeral key and the session key of an initiator complete a Machine-In-The-Middle attack where the initiator believes to be talking to A while the attacker impersonated I to some Responder R.  We provide a the bottom of this issue a picture describing the attack on a simplified, abstracted LAKE-EDHOC execution for clarity.

Possible actions:
 * mention that method 0 is better suited for TEE implementations, and that other methods offer weaker authentication guarantees
 * update the protocol to allow DH authentication with a TEE, where the session key would not allow to recompute a MAC. The TEE interface would then contain both an export function to compute PRK_4x3m as well as a function that would return the full MAC. Two similar options could be considered:
     - Option 1) Add to the MAC payload the value G^IY. This option does not update the key derivation.
	 - Option 2) Distinguish the session key and the MAC key, with an additional inclusion of the G^IY value for the MAC key. 

## Impersonation attack without compromise of long term authentication key
On an abstracted edhoc protocol:
 ![image](https://user-images.githubusercontent.com/7815574/161232673-19b4aa2a-f478-40e9-8e01-cb60bc2d1312.png)

gselander:
@charlie-j Thanks for presenting both problem and solution! 

Just for clarity (just reiterating text in #276): With PR #276  and by computing both the MAC and PRK_out inside a TEE, an attacker  with a leaked PRK_out would not be able to forge the MAC - thus strengthening the role of a TEE also for other methods than 0, right?

gselander:
Just a note about this:

> using a TEE for the long-term DH share does not currently have any meaningful impact on the security of the protocol


We found this statement quite strong :-) 

[...]

Issue #269, opened 2022-04-01
Simplify and improve the session key management


charlie-j:
As is, the session key PRK_4x3m offers much weaker properties than
the final exported keys.

Reason 1: session keys don't have "contributiveness" or "key control": even if Alice produces good randomness each time, keys may not be unique (if Bob sends low order points or the identity point - see footnote 1). Exported keys fix this by inclusion of transcript, which also gives a better binding to the protocol execution.

Reason 2: `PRK_4x3m` does not authenticate `TH_4` as it is independent of it. If there are ways "to play with" `MESSAGE_3` in order to obtain distinct values of `TH_4` on both sides (we discuss this further below), any application key derivation not using `TH_4` and only `PRK_4x3m` will break the authentication claim over `TH_4`.

As an additional remark, the current construction makes it difficult to identify in cryptographic terms what is the session key of the protocol, as the stored key material at the end is both `TH_4` and `PRK_4x3m`.

Our suggestion is to follow a design similar to e.g. TLS, and to have a final key derivation, that combines both `TH_4` and `PRK_4x3m` into a final key. This has multiple advantages:
  * solve the issues raise before, ensuring contributivity and no key control over the session key, as well as ensuring that the session key is tied to the execution through `PRK_4x3m`;
  * simplify the final state of the protocol, where instead of managing both `TH_4` and `PRK_4x3m`, only the final key is stored;
  * simplify the exporter, as it would only now depend on the session key.

To make this proposal a bit more concrete, we provide a first draft of a commit that could implement this solution here: https://github.com/charlie-j/edhoc/commit/9f4843a8ff9f93a0522ff3f1a478657fd197b5e2

## Breaking TH_4 authentication

We discuss here some subtle ways that may lead to a breach of authentication for `TH_4`, which could make the authentication very implementation dependent. Those mostly rely on the fact that using `ciphertext_3` inside the transcript instead of the underlying payload may allow the attacker some leeway in changing the concrete value of `ciphertext_3`, thus leading to an initiator and a responder disagreeing on the vlaue of `TH_4` after verification of message 3.

[...]

PR #276, opened 2022-04-07
Add a final key derivation

charlie-j:
This pull request corresponds to #269.

Changes:
 * `PRK_4x3m` is now `PRK_3m` only, i.e. only used for computing the MAC.
 * An additional key `PRK_out` is defined, which can be seen as the final session key, computed based on `PRK_3m` and `TH_4`.
 * The export now only depends on `PRK_out`. 
 
 Comments:
 * The name `PRK_out` is completely arbitrary.
 * The `TH_4` input that was inside the exporter has been replaced by the empty string. It is a shortcut that allows to use the existing syntax, but it is also arbitrary.
 * This update also corresponds to the option 2) of #270. Indeed, computing both the MAC and `PRK_out` could be performed inside a TEE, and a leak of `PRK_out` once it is outside of the TEE would not allow an attacker to forge any MAC it wants, as `PRK_3m` and `PRK_out` are now disjoint values.


emanjon (approved):

emanjon:
This looks like a good thing to do for several reasons. It also lowers memory requirements as the implementation only has to store PRK_out instead of (PRK_4x3m, TH_4).

emanjon:
`After verifying message_3, the Responder is assured that the Initiator has calculated the key PRK_3m (explicit key confirmation) and that no other party than the Initiator can compute the key.`

I assume that is enough. The document should probably say something about PRK_out as well. I assume the property we get it the the Responder knows that I can compute PRK_out.

emanjon (changes requested):
More text is needed on PRK_Out. What does R know about I and PRK_Out after receiveing flight 3 vs. after recieving flight 5. What does I know about R and PRK_Out after recieving flight 4.

charlie-j:
@emanjon you can check out the last update!

gselander:
I spotted a typo, otherwise the update looks ready to merge.

But following #278 there are quite a few other changes we want to make to the key schedule. To avoid merge conflicts should probably do the changes in sequence. We could start by merging this, although there will be further changes to some of the changes made here.

PR #282, opened 2022-04-22
Information leakage

emanjon:
- Rewrote the variable length information leakage section.

- PR needs to be augmented with information leakage from terminated EDHOC exchanges.


emanjon:
Merging this small correction for info laeakage.

Issue #278, opened 2022-04-17
Summarizing the the key derivation after the current suggested changes


emanjon:
There has been a lot of suggestions and discussion regarding the key derivation latelely. Below is an attempt to summarize all the current suggstions that currently seems relatively likely to be done. This is not an issue in itself but I think it is good to have somewhere to discucess the complete changes to the key derivation. 

-----------------------------------

- Suggested to permute the order of element in the TH_2 sequence to be more resistant to any future collision attacks in hash functions. Exact order unclear but suggested to put the random and unpredictable G_Y first.
- Suggested to use PLAINTEXT instead of CIPHERTEXT in transcript hashes to simplify formal verification. Endpoints agree on that they received the same plaintext rather then the same cipertext.

```
TH_2 = H( G_Y, C_R, H(message_1) )
TH_3 = H( TH_2, PLAINTEXT_2 ) 
TH_4 = H( TH_3, PLAINTEXT_3 )
```

- Suggested to use a salt derived from PRK in Extract to avoid usign the PRK keys directly in both Extract and Expand

```
PRK_3e2m = Extract( SALT_2e,   G_RX )
PRK_3m   = Extract( SALT_3e2m, G_IY )
```

- Suggested to use unsigned integer instead of text string for label to save memory.

```
OKM = EDHOC-KDF( PRK, transcript_hash, label, context, length )
    = Expand( PRK, info, length )

   info = (
     transcript_hash : bstr,
     label : uint,
     context : bstr,
     length : uint,
   )
```

- Suggestion to add a final key derivation PRK_out and use that in the exporter. Note that the exporter would also have uint labels

```
EDHOC-Exporter( label, context, length ) = EDHOC-KDF( PRK_out, h'', label, context, length )
```

- Suggestion to base Key-Update on Expand instead of extract. The label would need IANA registration. hash_length is a new variable that the implementation needs to keep track of. Should considered whether using Exporter for KeyUpdate has any disadvantages.
- Suggested that more info on how an application protocol is supposed to use KeyUpdate and Exporter together is needed.

```
PRK_out = EDHOC-KeyUpdate( context ) = EDHOC-Exporter( 0, context, hash_length )
```

- The EDHOC internal key derivations (i.e., not using the Exporter) would then be something like below. The exact labels should be discussed. Note that that only some of the derivations actually need a label to avoid collision

```
context_2 = << ID_CRED_R, CRED_R, ? EAD_2 >>
[...]

Issue #271, opened 2022-04-01
Future proofing the protocol: transcript collisions resistance


charlie-j:
Chosen prefix collisions are a particular kind of hash function collisions, where given two prefixes p1 and p2, the attacker is able to compute two bitstrings c1 and c2 such that `h(p1 | c1) = h(p2 | c2)`. The consequences of such collisions over protocols have been explored in [1]. Computing such collisions is feasible on MD5 and SHA-1, and it is unknown whether it will or not be the case for SHA-256 in the coming years. 

On the current version of the protocol, chosen prefix collisions lead to attacks on secrecy and authentication, as well as downgrade attacks on the ciphersuite. Such attacks can be prevented or mitigated by a set of small adjustment to the standard, and would at low cost improve the long term security of the protocol. Let us first detail the concrete chosen prefix collisions, before talking about the mitigations.

The attacker performs a Machine-In-The-Middle, where it will use a chosen-prefix collision attack to build a session where I and R are talking together and agree on the value of the hash of the transcript, but where each agent actually received the neutral DH group element (which is so that e^x=e).

The following transcripts are computed in such a run, where `Trans_E` is the expected shared transcript (displayed using variable names), and `Trans_I` and `Trans_R` are the transcripts produced by the attack, with values aligned with corresponding variable names of the expected transcript.
```
Trans_E :=  method |  suitesI  | G_X |  C_I  |  EAD_1  | G_Y | C_R 

Trans_I :=  zero   | "suitesI" | g^x | "C_I" | "EAD_1" |  e  | c2 | g^y | C_R 
Trans_R :=  zero   | "suitesI" |  e  | "C_I" |   c1    | g^y | "C_R"
```

We see that the attacker is using the arbitrary long fields `EAD_1` and `C_R` to stuff the bitstring corresponding to the chosen-prefix collision. In this scenario, we remark that as the ciphersuites are part of the prefix, the attacker could also decide to show two distinct suitesI values to both participants, corresponding to a downgrade attack. As a last scenario, if the parsing of suitesI is loose, for instance accepting `suites = [ 2* int / btstr ] / int` instead of `suites = [ 2* int ] / int`, a value of the array can also be use to stuff the chosen-prefix collision bits.

Possible actions:
 * add length restrictions over the `EAD,C_R` and `C_I` fields;
 * enforce/recommend checking for low-order points when it has a low-ressource cost (and in general at least the identity element);
 * enforce/recommend that the CBOR parsing must be strict.

[...]

PR #283, opened 2022-04-22
Change the input order of TH_2

charlie-j:
This allows to future proof the protocol against a weak hash with
chose-prefix collisions.

It corresponds to #271, and is also mentioned in #278.


emanjon (approved):
Looks good. I think should do this change.

Issue #273, opened 2022-04-01
Clarify private key usages


charlie-j:
The standard does not restrict private key usages. In particular, it could be tempting to use the public/private keys for the ed25519 signature scheme as a public and private key pair for the static authentication mode with the elliptic curve X25519. This does not have clear direct and practical implications. However, this usage falls outside the existing cryptanalysis for the signature scheme, and is a bad practice from a theoretical point of view. This would require a dedicated cryptographic analysis, similar to [2].

[1, Section 12] typically recommends that different keys are used for different algorithms, but it may be unclear if the EDHOC use case falls under this recommandation, or if an extra one could be pertinent. 

We recommend to mention this issue somewhere in the standard, or explicitly forbid such a shared key usage.
[1]: https://www.ietf.org/archive/id/draft-ietf-cose-rfc8152bis-struct-15.txt
[2]: https://eprint.iacr.org/2021/509.pdf On using the same key pair for Ed25519 and an X25519 based KEM. Erik Thormarker.

chrysn:
I've seen single keys used for both ed25519 and x25519 during plugtests; IIUC the agreement among the plugtesters back then was that this was not intended use and should generally not be done. This might serve as a data point that illustrates the need for a clear statement.

emanjon:
Mentioning this issue and refering to [1, Section 12] seems like a good idea. The text would need to mention EDHOC methods as [1, Section 12] only talks about algorithms and the ECDHE in EDHOC is not strictly a COSE algorithm. Writing that using the same key requires a dedicated cryptographic analysis, similar to [2] also seems like a good idea.

(As a comparision, I don't think TLS 1.3 states anything about forbidding the public key in the certificate to be used for Elgamal, ECIES, or ECDHE outside of TLS).

charlie-j:
Should we propose a PR for this?

gselander:
@charlie-j Please go ahead. Prepare for the option to have a reference to an analysis showing this is secure under certain conditions.

PR #284, opened 2022-04-27
Add recommendations for static key usages.

charlie-j:
A static key should not be used both for X25519 and Ed25519.

This PR corresponds to #273.


gselander:
Please review the proposed change.

Issue #281, opened 2022-04-20
Add that EDHOC uses a deterministic encoding.


emanjon:
It is stated that EDHOC uses deterministic CBOC, but EDHOC could still be non-deterministic if is allows a choice between e.g., bstr and int.

1. Add that the encodig in EDHOC is deterministic.

2. Add some considerations regarding decoding. In any cryptographic protocols the optimal for security is that the decoder raises a fatal error if the something in the encoding is enexpected. E.g. wrong order in map or two byte encoding of the integer 7. This is however not supported by CBOR implementations so it cannot be mandated. The security issues with a allowing decoder is that is gives options to an attacker. If the hash function has some weakness, An attacker can e.g. try all n! permutations of a map with n elements until they find a collision. This is not critical but it might be guidance that a decoder that only tolerated deterministic CBOR might have security benefits.

charlie-j:
That would be interesting, notably as it was mentioned as an attack surface previously (#269).

Bellow is an additional example of the kind of weird things that can go on when everything is not nice and deterministic.

# Privacy compromise against active attacker from R to I but passive from I to R

Consider an attacker that can only modify messages going from Rresponder R to Initiator I, but not send its own messages to R. Then, the identity of R should intuitively be anonymous. This could be the case of an attacker only sitting next to the device I, and who wants to know from whom I is receiving messages.
 
The attacker wants to tests if some I is communicating with some fixed identity `id_r`. This attacker may intercept the message_2 encrypted part: `enc := <id,signature,ead_2> XOR keystream`. Now assume that there exists a second distinct encoding of `id_r`, that we call `id_r2`. The attacker can then compute `<id_r,0,0> XOR <id_r2,0,0> XOR enc`.

Now, two options, either `id = id_r`, in which case the message received by I is `<id_r2,signature,ead_2> XOR keystream`, which will successfully proceed. If `id <> id_r`, the processing of the message will fail. And thus, the attacker can learn if I is talking to `id_r` or not.

Remark that in both cases, the key exchange will fail at the next step as parties will disagree on `TH_3`.

gselander:
The text about decoding should recommend to abort if encoding not deterministic.

PR #287, opened 2022-05-11
Deterministic encoding and decoding. CDDL

emanjon:
#281


gselander (commented):
"All EDHOC messages are deterministically encoded CBOR Sequences {{RFC8742}}."

Do we need to clarify the meaning of "deterministically encoded CBOR Sequences"? Readers may be lead to believe it is defined in RFC8742.

Issue #290, opened 2022-05-13
State max size of message_2 when SHA-2 is used


gselander:
Channeling from Erik.

Use of the HKDF as EDHOC-KDF for generation of KEYSTREAM_2 and encryption of message_2 induces a max size. For example, HKDF-SHA256 has max output size 255*32 = 8160 bytes.

PR #292, opened 2022-05-16
Cryptographic things to fix before submission of -14

emanjon:
- Reformulating security of expand as stream cipher instead of removing.
- HKDF-Expand is very often used.

- Added note of HKDF-Expand length restriction.

- Key use for different algorithms is already covered by COSE.
- Nothing is special with use inside of EDHOC.
- The text is clearly not correct at as signate can also be used in other methods.


emanjon:
#290

Issue #295, opened 2022-06-02
Connection identifers and privacy


emanjon:
If connection identifiers are predictable (e.g., a sequence) they might be used by an attacker to determine that the two EDHOC sessions came from the same device. We should add a sentence on this. In practice it does not matter unless IP address, MAC address, etc are randomized as well.

charlie-j:
Isn´t it covered by sec 8.5?: 
>    An attacker observing network traffic may use connection identifiers sent in clear in EDHOC or the subsequent application protocol to  correlate packets sent on different paths or at different times.

Essentially, any of the additional data (EADs and C_I/C_R) can be used to attack privacy, whether anonymity or unlinkability. A side remark is that even with padding, if any of the  EAD_s is a counter, the length of message_x can allow to distinguish high intensity traffic vs low intensity traffic.

emanjon:
Yes, it does. The text is more general then I thought.

```
   An attacker observing network traffic may use connection identifiers
   sent in clear in EDHOC or the subsequent application protocol to
   correlate packets sent on different paths or at different times.  The
   attacker may use this information for traffic flow analysis or to
   track an endpoint.  Application protocols using connection
   identifiers from EDHOC SHOULD provide mechanisms to update the
   connection identifier and MAY provide mechanisms to issue several
   simultaneously active connection identifiers.  See [RFC9000] for a
   non-constrained example of such mechanisms.
```

Should probablt have some guidance on non-predictable connection identifiers. If a device is using a sequence like 0x3489f2, 0x3489f3, 0x3489f4, 0x3489f5 updating the connection identifier is quite useless.

PR #298, opened 2022-06-03
Connection identifier privacy

emanjon:
#295

Some additional text on connection id privacy
- New connections identifiers needs to be unpredictable to provide improved privacy
- QUIC requirement to not use the same identifier in a connection seems like nonsense that lowers privacy compared to random selection and increases implementation complexity.
- Christian Amsuss pointed out in a discussion in CORE that connection id privacy often does not make sense in many deployments as there are fixed identifier in lower layers


gselander:
LGTM

Issue #300, opened 2022-06-07
Do we need different names for paddings of different messages?


gselander:
If there is a risk someone thinks that padding in the messages must be the same we could use   PAD_2, PAD_3, PAD_4.

charlie-j:
If I understand correctly, the PAD is to hide the length of the messages. If the three messages share the same PAD length, it is enough for one of the message to have a length known by the attacker, e.g. if EAD_4 is of constant length across all sessions, to leak the length of the PAD, and thus the length of the other messages contents. As such, 3 distinct padding lengths would be strictly better.

Otherwise, as the encryptions are distinct (different key streams), I on my side don't see any weird cryptographic interactions that would be possible.

(a side question is does the padding in message 2 leads to potential for padding oracles attacks, where the attacker could try to flip a few bytes of the cyphertext and learn some information based on which step of the processing fail. However, that is covered by #294 I guess)

emanjon:
This is an editorial change. I merged the PR and will close this issue.

`does the padding in message 2 leads to potential for padding oracles attacks, where the attacker could try to flip a few bytes of the cyphertext and learn some information based on which step of the processing fail.`

No and yes. SIGMA and EDHOC and TLS 1.3 does not provide confidentiality of the second flight to active attackers (except maybe the signature). In SIGMA and EDHOC and attacker can modify the second flight and gain information. But this attack seems very uninteresting compared to attacks where the attacker modifies message_1.

PR #302, opened 2022-06-09
Distinct padding names for each message

gselander:
#300


emanjon:
This is an editorial change.

PR #307, opened 2022-06-27
Unauthenticated Operation (TOFU)

emanjon:
- I think it would be good to point out early that the EDHOC protocol (just like TLS) only provides proof-of-possesion. Authentication has to be done by the application. I have seen horrible uses of TLS where the implementors believe RFC 8446 when it says it provides authentication.

- Added text for "Unauthenticated Operation". As pointed out by Marco, this is more general than just TOFU. "Unauthenticated operation" is the term used by TLS 1.3


gselander (commented):
I made some minor updates, have a look.

gselander:
I think this is ready to merge.

Issue #275, opened 2022-04-06
Encrypted EAD_1 (or general steps for enabling it)


chrysn:
This is motivated by two considerations, the concerns for clients selecting a hidden service in a protected way (pinging @eighthave, do you have a pointer to your hackathon project), and TLS's encrypted Client Hello (superseding eSNI, encrypted server name indication).

The rough situation for either case is that a client has some out-of-band information about the server that includes a public key for encryption (even if the encryption key on its own may be insufficient to authenticate the server).

As a straw man proposal (mainly as scaffolding for the "how to enable it" steps later), consider the following:

> ead_label -1 is defined as "Encrypted EAD", its corresponding ead_value is a byte string. To produce valid data in this option, the initiator has to know some public key G_E (for "early") to which the responder presumably has the private key E, and some EAD_1 items [possibly following the same shape as EAD in general] encoded into a plain text byte string. The byte string is produced by the client by obtaining a shared secret G_XE, and expanding it [...] to the length of the plain text, and XOR-ing them into the ead_value.
> [We might also describe that there is some inner MAC on the plaintext -- but it is important that anything outside visible is indistinguishable from random data].
>
> The responder may use plaintext recovered from reversing the encryption procedure (using its G_X and E rather than X and G_E to obtain the shared secret) for selecting the identity (CRED_R / ID_CRED_R) it presents to the client; the presence of the plain text MUST NOT have an impact on the authorization associated with the exchange.

(I bet that reading the [Encrypted Client Hello draft](https://datatracker.ietf.org/doc/draft-ietf-tls-esni/) will give valuable input for making this complete).

To ensure that responders (who are, in their identity selection, the more exposed side) can't be probed for their support of encrypted EAD, encrypted EAD should be marked as "elective" in the sense that a server may ignore it if it does not support it. (It'd still become part of the transcript). If a server receives an EAD it can not decrypt (or which, after decryption, contains nothing valid, eg. because the wrong G_E was used), it needs to just continue and pick the CRED_R / ID_CRED_R it would always have picked.

[...]

PR #311, opened 2022-06-28
 Updated EAD processing (critical and non-critical)

gselander:
Addressing #275

### DIFF
--- a/draft-ietf-lake-edhoc.md
+++ b/draft-ietf-lake-edhoc.md
@@ -576,7 +576,7 @@ In order to reduce round trips and the number of messages, or to simplify proces
 
 EDHOC allows opaque external authorization data (EAD) to be defined in a separate specification, and sent in dedicated fields of the four EDHOC messages (EAD_1, EAD_2, EAD_3, EAD_4).
 
-Each EAD field is a CBOR sequence (see {{CBOR}}) consisting of one or more (ead_label, ead_value) pairs as defined below:
+Each EAD field is a CBOR sequence (see {{CBOR}}) consisting of one or more EAD items (ead_label, ead_value) as defined below:
 
 ~~~~~~~~~~~ CDDL
 ead = 1* (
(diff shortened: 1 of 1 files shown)


emanjon:
Issue needed.

draft-ietf-lake-edhoc.md
```diff
@@ -585,7 +585,11 @@ ead = 1* (
 )
 ~~~~~~~~~~~
 
-A security application using external authorization data need to register an ead_label, specify the ead_value format for each message (see {{iana-ead}}), and describe processing and security considerations.
+A security application using external authorization data need to register an ead_label, specify the ead_value formatting for each message (see {{iana-ead}}), and describe processing and security considerations.
+
+An EAD field can be either critical or non-critical, determined by the sign of ead_label: negative means critical; non-negative means non-critical. If an endpoint does not support a received critical EAD field, the EDHOC protocol MUST be discontinued.
```
emanjon, line 590: "If an endpoint does not support a received critical EAD field, the EDHOC protocol MUST be discontinued."

This needs more discussion and work. The handling of critical extensions in C.509 is to my understanding confusing and problematic. Both in the  specification and in implementations.

What does support means. Can an endpoint supporting the EAD ignore it.

emanjon (commented):
I think critical ead is a very good discussion to have. I would suggest that we have a discussion and add any outcome to -16 shortly after IETF 114.

draft-ietf-lake-edhoc.md
```diff
@@ -585,7 +585,11 @@ ead = 1* (
...
+
+The EAD specification describes if an EAD field is always critical, always non-critical, or if it can be decided by the application in the sending endpoint. In the latter case, one negative and one non-negative instance of ead_label need to be registered.
```
emanjon, line 592: The specification likely MUST describe

draft-ietf-lake-edhoc.md
```diff
@@ -1465,7 +1469,7 @@ IANA has created a new registry entitled "EDHOC Error Codes" under the new group
 
 ## EDHOC External Authorization Data Registry {#iana-ead}
 
-IANA has created a new registry entitled "EDHOC External Authorization Data" under the new group name "Ephemeral Diffie-Hellman Over COSE (EDHOC)". The registration procedure is "Specification Required". The columns of the registry are Label, Message, Description, and Reference, where Label is an integer and the other columns are text strings.
+IANA has created a new registry entitled "EDHOC External Authorization Data" under the new group name "Ephemeral Diffie-Hellman Over COSE (EDHOC)". The registration procedure is "Specification Required". The columns of the registry are Label, Description, and Reference, where Label is an integer and the other columns are text strings.
```
emanjon, line 1472: This should likely be non-negative integer now.

emanjon: I did some changes to the registration procedure and the processing regarding critical EAD fields. I think the current version could be included in -15

gselander:
Should we allow or forbid that ead_label  = k and ead_label = -k correspond to different ead-values?

For example, if an (ead_label, ead_value) is always critical: 

* Say, ead_value is an ACE Access Token with ead_label = -4, always critical.

Then we can either:
1. forbid the use of ead_label = 4 altogether, or 
2. allocate an always non-critical EAD to ead_label = 4, say, ead_value is "service indication" as in #275

In case 1 we will have some "holes" in the register. One positive integer blocked for each always critical (ead_label, ead_value) , one negative blocked for each always non-critical (ead_label, ead_value), and 0 cannot be used. (It is not clear to me that there are many always non-critical, but always critical is probably common.)

In case 2, we could still have the registration policy that: if an EAD can be either critical or non-critical then two ead_label with same absolute value should be reserved for it.  

What is the main risk with case 2? If someone misunderstands or by accident changes sign on an always /non-/critical EAD this results most likely in an error due to wrong ead_value (like in the example above), but potentially in an unintended EAD message. (Another registration policy could be that the ead_values corresponding to ead_labels k and -k should have incompatible CDDL so the error is discovered ...) The EAD specification needs anyway to specify what are the processing rules and if you violate those more or less anything could happen. 

Comments?

gselander:
BTW, we may want a term for the ordered pair ead_label, ead_value). The term "EAD" or "EAD field" is referring to the message field, which may consist of multiple (ead_label, ead_value).

emanjon:
> What is the main risk with case 2?

Just seems like overoptimization to me.

gselander:
Wait with this pending design team meeting, see #275.

gselander:
I made a proposal. The text now include the concepts "EAD item" and "critical"/"non-critical" EAD item. There is also more details about how applications makes use of EAD: 
* each application registers its own ead_labels
* ead_label is a positive integer associated to a particular ead_value
* if the application transports different ead_values then multiple ead_labels need to be registered
* the application specifies how EAD items are transported in EDHOC messages, a particular EAD item may be transported in multiple messages (i.e. different EAD_x).

I'll merge this into -15 and look forward to comments!

PR #315, opened 2022-07-05
Vectors Updated to spec 14

stoprocent:
Hi
Sending updated vectors to spec 14. For now only p256 but yet to be verified. 

Code to generte can be found here: 
https://github.com/stoprocent/edhoc/commit/68f93cd3884d3ff232acf0466ea0d3e8014936d7


gselander:
Thanks! This is updating test-vectors-11/p256/. Do we want to overwrite those? What do you think about instead making a new directory test-vectors-14/p256/ ?

gselander:
Or maybe keep the directory. I'm submitting traces-01 now which states:

"Editor's note: update reference to test vectors below:

A more extensive test vector suite and related code that was used to generate trace 1 can be found at: https://github.com/lake-wg/edhoc/tree/master/test-vectors-11. The test vector for trace 2 can be found at: https://github.com/lake-wg/edhoc/tree/master/test-vectors-11/p256."

Issue #324, opened 2022-07-27
Computational analysis of EDHOC STAT - improvement suggestions for K_3


gselander:
Proposal to derive K_3 from PRK_4e3m instead of PRK_3e2m, to make K_3 dependent on the static key of the Initiator.

charlie-j:
Wait... Isn' t that impossible?
The responder needs to be able to decrypt message 3 in order to know G_I, and G_I is needed to compute PRK_4e3m. So the key K_3 used to decrypt message 3 cannot depend on G_I.
Am I missing something?

gselander:
I had the same comment. Please see slides 21ff in 
https://datatracker.ietf.org/meeting/114/materials/slides-114-lake-computational-analysis-of-edhoc-stat-stat-00

My interpretation was to partition PLAINTEXT_3 into, say,

PLAINTEXT_3A =
(
? PAD_3,
ID_CRED_I / bstr / -24..23,
? EAD_3
)

PLAINTEXT_3B = Signature_or_MAC_3

KEYSTREAM_3A = EDHOC-KDF( PRK_3e2m,   TBD, TH_3, plaintext_length )
K_3B         = EDHOC-KDF( PRK_4e3m,   TBD, TH_3,      key_length )
IV_3B        = EDHOC-KDF( PRK_4e3m,   TBD, TH_3,      iv_length )

CIPHERTEXT_3A = PLAINTEXT_3A XOR KEYSTREAM_3A   

CIPHERTEXT_3B = AEAD encryption of PLAINTEXT_3B as is currently done in message_3, with K_3B/IV_3B and
additional external_aad = PLAINTEXT_3A.

CIPHERTEXT_3 is the concatenation of CIPHERTEXT_3A and CIPHERTEXT_3B.

This would increase the size of message_3 with 1-2 bytes due to separate CBOR encoding of CIPHERTEXT_3A and CIPHERTEXT_3B.

I'm not sure this solves the problem though.

emanjon:
Some quick thought on the PR and alternatives

OLD (draft-ietf-lake-edhoc-15)
AEAD ( PAD_3, ID_CRED_I, Signature_or_MAC_3, EAD_3 )

ALT 1 (#325)
ENC( PAD_3, ID_CRED_I, EAD_3 ) || AEAD( Signature_or_MAC_3 )

ALT 2 (Seems preferable to #325 )
[...]

PR #325, opened 2022-07-28
New encryption of message_3

gselander:
#324

### DIFF
--- a/draft-ietf-lake-edhoc.md
+++ b/draft-ietf-lake-edhoc.md
@@ -752,16 +752,16 @@ Further details of the key derivation and how the output keying material is used
 
 ~~~~~~~~~~~~~~~~~~~~~~~
 KEYSTREAM_2   = EDHOC-KDF( PRK_2e,   0, TH_2,      plaintext_length )
-KEYSTREAM_3A  = EDHOC-KDF( PRK_3e2m, TBD, TH_3,      plaintext_length )
-SALT_3e2m     = EDHOC-KDF( PRK_2e,   TBD, TH_2,      hash_length )
-SALT_4e3m     = EDHOC-KDF( PRK_3e2m, TBD, TH_3,      hash_length )
-K_3B          = EDHOC-KDF( PRK_4e3m, TBD, TH_3,      key_length )
-IV_3B         = EDHOC-KDF( PRK_4e3m, TBD, TH_3,      iv_length )
-MAC_2         = EDHOC-KDF( PRK_3e2m, TBD, context_2, mac_length_2 )
-MAC_3         = EDHOC-KDF( PRK_4e3m, TBD, context_3, mac_length_3 )
-PRK_out       = EDHOC-KDF( PRK_4e3m, TBD, TH_4,      hash_length )
-K_4           = EDHOC-KDF( PRK_4e3m, TBD, TH_4,      key_length )
-IV_4          = EDHOC-KDF( PRK_4e3m, TBD, TH_4,      iv_length )
+SALT_3e2m     = EDHOC-KDF( PRK_2e,   1, TH_2,      hash_length )
+MAC_2         = EDHOC-KDF( PRK_3e2m, 2, context_2, mac_length_2 )
+KEYSTREAM_3A  = EDHOC-KDF( PRK_3e2m, TBD, TH_3,    plaintext_length )
+K_3B          = EDHOC-KDF( PRK_4e3m, 3, TH_3,      key_length )
+IV_3B         = EDHOC-KDF( PRK_4e3m, 4, TH_3,      iv_length )
+SALT_4e3m     = EDHOC-KDF( PRK_3e2m, 5, TH_3,      hash_length )
+MAC_3         = EDHOC-KDF( PRK_4e3m, 6, context_3, mac_length_3 )
+PRK_out       = EDHOC-KDF( PRK_4e3m, 7, TH_4,      hash_length )
+K_4           = EDHOC-KDF( PRK_4e3m, 8, TH_4,      key_length )
+IV_4          = EDHOC-KDF( PRK_4e3m, 9, TH_4,      iv_length )
 ~~~~~~~~~~~~~~~~~~~~~~~
 {: #fig-edhoc-kdf title="Key derivations using EDHOC-KDF."}
 {: artwork-align="center"}
--- a/draft-ietf-lake-edhoc.md
+++ b/draft-ietf-lake-edhoc.md
@@ -752,15 +752,16 @@ Further details of the key derivation and how the output keying material is used
 
 ~~~~~~~~~~~~~~~~~~~~~~~
 KEYSTREAM_2   = EDHOC-KDF( PRK_2e,   0, TH_2,      plaintext_length )
-SALT_3e2m     = EDHOC-KDF( PRK_2e,   1, TH_2,      hash_length )
-MAC_2         = EDHOC-KDF( PRK_3e2m, 2, context_2, mac_length_2 )
-K_3           = EDHOC-KDF( PRK_3e2m, 3, TH_3,      key_length )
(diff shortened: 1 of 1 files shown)


gselander:
This is still experimental. Comments from implementers and reviewers are most welcome!

@charlie-j @marcilunga Can you say anything how this impact your analysis?

gselander:
Sorry, I made unnecessary mess of the key derivation table. Hang on for a short while and I'll make the diff more readable!

gselander:
Now better (when using "Files changed" above)

charlie-j:
I am unavailable in August, so formal results will need to wait for September.
At a first glance, it looks like we are strengthening the encryption of the signature, but weakening the one of the identity, because it is now malleable and not integrity protected. So, I am unclear if this is or not an improvement overall, and this may lead to unexpected privacy breaches.

What looks weird is that we are weakening the encryption of the identity of the initiator, in order to strengthen the encryption of the MAC, but the MAC should be a useless random value anyway. Is it really that worthwhile to be able to encrypt the MAC, based on PK_4e3m, with a key also based on PK_4e3m?

gselander:
@charlie-j Thanks for good comments. For clarity: this change is not decided, but for discussion. 

About 

> weakening the one of the identity, because it is now malleable and not integrity protected. 

Clearly CIPHERTEXT_3A does not integrity protect ID_CRED_R. But wouldn't appropriate use of external_aad in the COSE_Encrypt0 of CIPHERTEXT_3B  provide integrity protection in the same way as CIPHERTEXT_3 in -15? For example by including PLAINTEXT_3A in external_aad?

gselander:
@charlie-j Also, thanks for sharing your timeplan!

charlie-j:
The content of plaintext_3a would indeed be integrity protected by the later operations.  But in a sense, here we are emulating an AEAD by splitting up into two parts the decryption and the integrity check. This is often error prone, and what paved the way for some padding oracle attacks or timing attacks.

A possible weird behavior:
1) Assume two ID_CRED_I and ID_CRED_I' points  to the same  long term g_I (is it possible?) 
 2) During the third message between an anonymous initiator with ID_CRED_X talking to an honest responder, an active attacker intercepts it and xor the ID_CRED_X part with (ID_CRED_I xor ID_CRED_I'). Then, two possibilities:
  2.1) ID_CRED_X is not I or I', and thus, the responder will either not find an ID_CRED corresponding to (ID_CRED_X xor ID_CRED_I xor ID_CRED_I'), or use some unrelated g_X, in both cases, decryption of cyphertextB will fail 
  2.2) ID_CRED_X is I or I', then we have that the responder actually receives either ID_CRED_I or ID_CRED_I', the responder then uses g_I and decryption will succeed. However, the signature check will fail.
  
 On most implementations, it would be easy to time for an attacker the failure point, and thus know whether the decryption or the signature failed.  And in the previous scenario, the attacker can then know whether 2.1 or 2.2 happened, and thus whether the initiator is I. Thus, an active attacker could potentially break the anonymity of the initiator.

baptistecottier:
> What looks weird is that we are weakening the encryption of the identity of the initiator, in order to strengthen the encryption of the MAC, but the MAC should be a useless random value anyway. Is it really that worthwhile to be able to encrypt the MAC, based on PK_4e3m, with a key also based on PK_4e3m?

In fact, with the actual protocol, message_3 is not authenticated:
The key K_3 used for the encryption in the message 3 depends only of the initiator ephemeral key (X), the responder long-term key (R) and the responder ephemeral key (Y). So an adversary can compute the key K_3 and therefore, this is not an authenticated encryption. 

The idea in this request is to compute an AEAD with the key K_4 that depends on the Initiator long-term secret key I. However, the responder need credential informations from the Initiator to compute K_4. Those informations are encrypted under K_3 as the first part of message_3 that the responder will decrypt before computing the key K_4.

> an active attacker intercepts it and xor the ID_CRED_X part with (ID_CRED_I xor ID_CRED_I')

We can also imagine xoring the ID_CRED_X part with (ID_CRED_I1 xor ID_CRED_I2) allowing the adversary to test two identities in the same time. This malleability indeed seems to be a weakness of this update. Considering a classic block cipher encryption with K_3 would solve this weakness but at the price of a larger message_3 (I guess).

chrysn:
If some part of message_3 gets just encrypted and not immediately authenticated (eg. through a KEYSTREAM_3), I think this opens up a (very narrow) attack from a MITM attacker:

Assume (and yes that's a lot of assumptions, maybe some can be dropped) an attacker knows that ID_CRED_I are of a very particular shape (say, all in the 1-byte form), that no padding or EAD_3 is used. Further assume that the responder's behavior differs between "ID_CRED_I is not recognized" and "The validation failed" (even if it's just the timing -- other behavioral differences can be forbidden, but not leaking timing would be hard. After all, that's why we use AEADs in the first place).

Now assume that the attacker sees repeated EDHOC exchanges between the responder and what is apparently the same initiator. Then the attacker can
* start flipping bits in where it knows the ID_CRED_I to be, and thus count how many ID_CRED_I values are recognized by the server, or alternatively
* knowing that there are only two ID_CRED_I and knowing one ID_CRED_I that is usable, vary the ID_CRED_I in message_3 and thus determine which ID_CRED_I the attacked client has.
  (More elaborate scenarios are probably possible with multiple initiators, jumping from one to the next).

Thus, I'd be very cautious about non-AEAD encryption in message 3.

(I'm already unhappy about having to parse unauthenticated data in message 2, but there it's about ID_CRED_R, and EDHOC does not claim to protected that ... which gives me input for encrypted client hello at #275).

emanjon:
`The key K_3 used for the encryption in the message 3 depends only of the initiator ephemeral key (X), the responder long-term key (R) and the responder ephemeral key (Y).`

What do you mean with an attacker here? Any entity can of course take the role R, whether I wants to talk to R is determined in I's processing of message_3. But is Alice and Bob exchange message_1 and message_2 an adversary Eve cannot create a valid message_3 and R is assured that only I can read message_3 including ID_CRED_I.... or?

baptistecottier:
@emanjon wrote:
> What do you mean with an attacker here? Any entity can of course take the role R, whether I wants to talk to R is determined in I's processing of message_3. But is Alice and Bob exchange message_1 and message_2 an adversary Eve cannot create a valid message_3 and R is assured that only I can read message_3 including ID_CRED_I.... or?

In what I said, the attacker would be the Initiator. The only key that authenticates the Initiator is I. And this key is implied in the key schedule after the computation of K_3 and IV_3. So message_3 can not an authenticated ciphertext, as the key K_3 used is not authenticated.

baptistecottier:
In fact, one can imagine that rather than stopping the protocol in case ID_CRED_I is not recognized or the verification fails, the responder chose a random credential, and keep following the protocol, avoiding some timing attack from an eavesdropper. Therefore the modifications made in this PR may work.

chrysn:
On Mon, Aug 01, 2022 at 03:41:03AM -0700, Baptiste Cottier wrote:
> [...]  in case ID_CRED_I is not recognized or the verification fails,
> the responder chose a random credential, and keep following the
> protocol, avoiding some timing attack from an eavesdropper

I doubt that this is a practical choice for implementers. Leaving out
the difficulties around random numbers (taking "the first" credential
would be good enough), this is putting extra processing in a place that
is still security critical. It would be making requirements of
no-side-channels that are not otherwise required of EDHOC
implementations. (AEAD decryption may have these requirements, but
that's the very point of authenticated encryption: do these properly,
once and for all).

If we had primitives for the sequence of decryption, then processing
putative plaintext, and then verification in COSE, and a good
understanding on what ill-effects it can have and what best practices
are, fine -- but we don't, and it shows by us coming up with surprise
leaks and hot-fixes in this thread. Researching such abstractions might
be a worthwhile endeavor, but I'd prefer not to have the exploration of
that as a dependency of EDHOC publication.

gselander:
Closing this, as agreed in the interim.

PR #337, opened 2022-09-29
Traces 03

marco-tiloca-sics:


draft-ietf-lake-traces.md
```diff
@@ -1790,35 +1789,35 @@ Then, PRK_2e is calculated using Extract() determined by the EDHOC hash algorith
     PRK_2e = Extract( salt, G_XY ) =
            = HMAC-SHA-256( salt, G_XY )
 
-where salt is the zero-length byte string:
+where salt is TH_2:
 
 ~~~~~~~~
 salt (Raw Value) (0 bytes)
 ~~~~~~~~
```
malishav, line 1796: should this be TH_2 here?

malishav: Question: Should TH_2 used as salt be the raw value of the cbor wrapping?

marco-tiloca-sics: > should this be TH_2 here?

Indeed. Now fixed, as well as the value of PRK_2e below. All should be ok after that.

marco-tiloca-sics: > Question: Should TH_2 used as salt be the raw value or the CBOR wrapping?

My understanding is that it should be the raw value, just like G_XY used as IKM parameter.

malishav: Thanks. With the fix above, I confirm the other values and my implementation should be ready for interop :)

gselander:
Merging in preparation for submission.

Issue #329, opened 2022-07-28
Comment from Marco to clarify identifier privacy recommendations.


emanjon:
Suggestion to clarify and maybe give example of randomness and unpredictability. The SHOULD should probably be removed. Unclear if 1 byte identifiers can ever by unpredictable.

PR #341, opened 2022-10-11
Issue #329 - Comment from Marco to clarify identifier privacy recommendations

emanjon:
#329


gselander (approved):
Looks good to me. @marco-tiloca-sics ?

draft-ietf-lake-edhoc.md
```diff
@@ -1287,7 +1287,7 @@ Symmetric algorithms used in EDHOC such as SHA-256 and AES-CCM-16-64-128 are pra
...
+An attacker observing network traffic may use connection identifiers sent in clear in EDHOC or the subsequent application protocol to correlate packets sent on different paths or at different times. The attacker may use this information for traffic flow analysis or to track an endpoint. Application protocols using connection identifiers from EDHOC SHOULD provide mechanisms to update the connection identifier and MAY provide mechanisms to issue several simultaneously active connection identifiers. See {{RFC9000}} for a non-constrained example of such mechanisms. Connection identifiers can e.g., be chosen randomly among the set of unused 1-byte connection identifiers. Connection identity privacy mechanisms are only useful when there are not fixed identifiers such as IP address or MAC address in the lower layers.
```
marco-tiloca-sics, line 1290: > chosen randomly among the set of unused 1-byte connection identifiers


To be sure, do you actually mean identifiers that are 1-byte in their native format, or instead that take 1 byte on the wire?

In the former case, that's up to 256 identifiers, but without including the 0-byte identifier and including mostly identifiers that take 2 bytes on the wire.

In the latter case, that's up to only 49 identifiers, but including also the 0-byte identifier and ensuring minimal size on the wire.

emanjon: It is an example, so does it matter?

marco-tiloca-sics: No, either case is fine; I just tend to get the first meaning from the new text. Anyway, it looks overall good to me :-)

Issue #340, opened 2022-10-04
Add text on properties of using several MACs


emanjon:
This is the outcome of the work by Baptiste Cottier and David Pointcheval related to #324

The conclusions form Baptiste and David is that after receiving message_4 the Initiator has authenticated the Responder with a strength of the combined MACs in message_2 and message_4. I.e. if both MACs are 64 bits, they work together as a 128 bit MAC.

The same thing apply to application messages such as OSCORE used as message 4,5,6,..... but this comes with the additional requirement that the security association has to be terminated and EDHOC redone in the case that verification of the OSCORE MAC fails. The same apply to the draft-ietf-core-oscore-edhoc where 2 MACs are sent in flight 3
https://datatracker.ietf.org/doc/draft-ietf-core-oscore-edhoc/

The same priciple should apply to modes with signatures where a signature in message_2 combined with a MAC in message_4 gives the strenght of the signature and the MAC combined. Baptiste and David have not investigated this formally.

PR #342, opened 2022-10-11
#340 - Add text on properties of using several MACs

emanjon:


gselander (approved):
Good. 

* Terminology comment:
   * You use first "authenticated with 128-bit security" and 
then "128-bit mutual authentication". I assume that means the same in security terms (although one is unilateral and the other mutual). 
   * No mentioning of online attacks here. Is it clear how that relates to 
"64-bit security against online brute force attacks" which is stated in another security consideration?

*  "can be achieved after verification of the first OSCORE Response". This is correct in its brevity, but the use of "verification" gives the impression that it is the verification of OSCORE Response is what makes this achieved, rather than it is the completion of the processing of the OSCORE request and OSCORE response that are the moments in time when, with the appropriate measures, security *can* be achieved. Perhaps replace "verification" with "completion".

gselander:
@baptistecottier Do you or David have any comment on the formulation?

gselander:
@baptistecottier We needed to merge this PR to prepare for -17, but we are of course still interested in your comments :-)

baptistecottier:
> @baptistecottier Do you or David have any comment on the formulation?

No comment to do ;). It's good.

PR #343, opened 2022-10-20
Add new formal analysis paper

charlie-j:
Hi all,

Our previous formal analysis on draft 12 and 14 lead to a publication at USENIX 23 available here https://hal.inria.fr/hal-03810102/.

As the two other previous formal analysis were cited in the draft, I'm proposing to add next to them our new one with this PR.

As a side remark, our models are here (https://github.com/charlie-j/edhoc-formal-analysis), and we are trying to update everything according to draft 17 with a new overall pass. It might be tight for IETF115, but we are trying our best! And anyway, it should only be positive results by now ;)

### DIFF
--- a/draft-ietf-lake-edhoc.md
+++ b/draft-ietf-lake-edhoc.md
@@ -168,6 +168,20 @@ informative:
         ins: NSA
     date: August 2015
 
+  Jacomme23:
+    target: https://hal.inria.fr/hal-03810102/
+    title: A comprehensive, formal and automated analysis of the EDHOC protocol
+    author:
+      -
+        ins: C. Jacomme
+      -
+        ins: E. Klein
+      -
+        ins: S. Kremer
+      -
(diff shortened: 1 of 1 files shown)


gselander:
@charlie-j Thanks! Great that you follow up with the latest version. The latest relevant changes are probably in -16 and only adds things to the key derivation (some of which was proposed by you :-).

charlie-j:
Thanks for the pointer! Though, the biggest issue is not so much updating the model of the protocol, but rather making a new full pass on the security claims and new appendices,  trying to have a clear systematized link between claims from the draft and what we check in the model.

emanjon:
@gselander let's check if there are some other papers missing and what we want to say about them.

gselander:
We should at least add references to work by ETH and ENS. Let's merge this as a starting point and continue discussion in #350.

Issue #348, opened 2022-11-02
Denial-of-service mitigation by increasing size of message_1


emanjon:
The Denial-of-Service text in EDHOC mentions Echo. Another way to mitigate Denial-of-service used by QUIC is to intentionally increase the size of message_1. Can be done by adding `? PAD_1` or by sending a dummy `EAD`. I think the document should at least mention the use of EAD_1 as a way to prevent Denial-of-Service.

emanjon:
```
   message_1 = (
     METHOD : int,
     SUITES_I : suites,
     G_X : bstr,
     C_I : bstr / -24..23,
     ? EAD_1 : ead,
   )
```

Three alternatives to make message_1 bigger

1. Add `? PAD_1 `
2. Add a lot of suites to `SUITES_I`
3. Add a padding `EAD_1 `

emanjon:
I think EAD_1 is a good way to go. I plan to make a PR discussing amplification and the solution of making message_1 bigger mentioning the EAD alternative.

emanjon:
If a EAP padding type would be registered, it could be argueed that ? PAD_2, ? PAD_3, and ? PAD_4 is not needed.

```
   ead = 1* (
     ead_label : int,
     ead_value : bstr,
   )
```

Note that with EAD, padding is minimum 2 bytes. I.e., 1 byte padding is not possible.

marco-tiloca-sics:
Just like for any other EAD items, if we use a new EAD_1 item for padding message_1, this item will also have to be handed off to "the application" for processing external to EDHOC, where it can be treated as padding and ignored.

In the other EDHOC messages, padding and its addition/removal does pertain to the core EDHOC message processing. That's why using PAD_1 can actually be better and more aligned with how padding is done already in other messages.

A good value for PAD_1 would be a sequence of CBOR simple value "true", like for the other PAD_X. A good position for it would be after C_I, i.e., before the EAD_1 field (if present).

emanjon:
Thanks for the quick input Marco. Let's discuss at IETF 115. Göran pointed out that this issue might be related to #275

gselander:
Unfortunately, we missed to discuss this in the LAKE meeting yesterday.

Anyone against having optional padding after C_1 in message_1?

Using the same encoding as for the other padding: PAD_1 = 1* true

emanjon:
This would need some discussion.

- Adding `PAD_1 = 1* true` is a change to the wire format
[...]

PR #357, opened 2022-11-09
Padding of message_1

gselander:
#348

### DIFF
--- a/draft-ietf-lake-edhoc.md
+++ b/draft-ietf-lake-edhoc.md
@@ -852,7 +852,7 @@ where:
 * SUITES_I - array of cipher suites which the Initiator supports in order of preference, the first cipher suite in network byte order is the most preferred by I, the last is the one selected by I for this session. If the most preferred cipher suite is selected then SUITES_I contains only that cipher suite and is encoded as an int. The processing steps are detailed below and in {{wrong-selected}}.
 * G_X - the ephemeral public key of the Initiator
 * C_I - variable length connection identifier. Note that connection identifiers are byte strings but certain values are represented as integers in the message, see {{bstr-repr}}.
-* PAD_1 - padding that may be used to increase message_1 for reducing amplification factor
+* PAD_1 - padding that may be used to reduce the amplification factor
 * EAD_1 - external authorization data, see {{AD}}.
 
 ### Initiator Processing of Message 1 {#init-proc-msg1}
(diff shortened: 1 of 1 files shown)


draft-ietf-lake-edhoc.md
```diff
@@ -839,6 +839,7 @@ message_1 = (
   SUITES_I : suites,
   G_X : bstr,
   C_I : bstr / -24..23,
+  ? PAD_1 : 1* true,
```
cabo, line 842: Please check CDDL

gselander: @cabo How about this?

? PAD_1 : padding,

...

padding = 1* true

cabo: 1* true is not a type, but a group.  So each true would turn up as a separate item in the sequence.
If that is OK, `* PAD_1: true` makes this more transparent.
If not, `? PAD_1: [* true]` is better.
(Note that I'm using * and not + (1*), because that allows more fine-grained control over the size of the padding.)

cabo: And don't forget to update the copy...

emanjon: ? PAD_1: [* true]
(this changes the wire format, and should probable by discussed with the implementors)

* PAD_1: true
(makes PAD_1 an item instead of a group, PAD_2, PAD_3, and PAD_4 are groups)

A third alternative would be:

message_1 = (
  METHOD : int,
  SUITES_I : suites,
  G_X : bstr,
  C_I : bstr / -24..23,
  ? PAD_1,
  ? EAD_1 : ead,
)

PAD_1 = 1* true

cabo: Of course, they should all be the same constructions (and, no, I don't want to change the wire format).
An optional sequence of at least one sounds a bit baroque.
Oh, and why not padding at the end?

gselander: What is Sturm und Drang with an optional sequence of at least one:
If there is none, why should there be a sequence?
Marco preferred this location of the padding, I think because it is the end of the simple EDHOC-internal processing; EAD_1 may contain multiple EAD items intended for the application.

cabo: What you find on the wire will be multiple `true` items at the sequence level.  Why not represent this with some fidelity?

gselander:
Closing this in favor of #360.

PR #358, opened 2022-11-10
Editorial: clarify the usage of C_X as OSCORE Sender/Recipient ID.

dnav:
Signed-off-by: David Navarro <david.navarro@ioterop.com>


gselander (approved):
@dnav Thanks for the table. Much clearer. I made some editorials, have a look if it is still OK.

emanjon (approved):

PR #360, opened 2022-11-11
EAD padding

emanjon:
#348


draft-ietf-lake-edhoc.md
```diff
@@ -1296,11 +1289,11 @@ An attacker observing network traffic may use connection identifiers sent in cle
...
+The fields ID_CRED_I, ID_CRED_R, EAD_2, EAD_3, and EAD_4 have variable length and information regarding the length may leak to an attacker. An passive attacker may e.g., be able to differentiating endpoints using identifiers of different length. To mitigate this information leakage an implementation may ensure that the fields have fixed length or use padding. An implementation may e.g., only use fix length identifiers like 'kid' of length 1. Alternatively padding may be used to hide the true length of e.g., certificates by value in 'x5chain' or 'c5c'. Padding in EDHOC is achieved by using the non-critical EAD item with EAD label TDB and a random bytestring as EAD value.
```
cabo, line 1292:
```suggestion
The fields ID_CRED_I, ID_CRED_R, EAD_2, EAD_3, and EAD_4 have variable length and information regarding the length may leak to an attacker. An passive attacker may e.g., be able to differentiating endpoints using identifiers of different length. To mitigate this information leakage an implementation may ensure that the fields have fixed length or use padding. An implementation may e.g., only use fix length identifiers like 'kid' of length 1. Alternatively padding may be used to hide the true length of e.g., certificates by value in 'x5chain' or 'c5c'. Padding in EDHOC is achieved by using the non-critical EAD item with EAD label TBD and a random bytestring as EAD value.
```

Nice!  But I think you mean TBD (To be defined), not TDB (threat database?).

cabo: Zero would be a good value for TBD.  (Would need to change some "positive" to "unsigned".)
(BTW: I'm not sure how you would register negative, i.e. critical, EAD label values?)

gselander:
Some updates. Open issues: 

* Still the option to do EAD_x = * ead, and make EAD fields mandatory.

* There is no way padding exactly 25 bytes, is there?

draft-ietf-lake-edhoc.md
gselander: I used zero. Why do we additional need text in 3.5?

marco-tiloca-sics:
> There is no way padding exactly 26 bytes, is there?

It's possible if we admit the padding EAD item to be repeatable.

In general, a specification defining an EAD item would also define if the item is repeatable.

draft-ietf-lake-edhoc.md
```diff
@@ -579,25 +579,40 @@ In order to reduce round trips and the number of messages, or to simplify proces
...
 
-An EAD item can be either critical or non-critical, determined by the sign of the ead_label in the transported EAD item included in the EDHOC message. Using the registered positive value indicates that the EAD item is non-critical. The corresponding negative value indicates that the EAD item is critical. ead_label = 0 MUST NOT be used.
+An EAD item can be either critical or non-critical, determined by the sign of the ead_label in the transported EAD item included in the EDHOC message. Using the registered positive value indicates that the EAD item is non-critical. The corresponding negative value indicates that this EAD item is critical. ead_label = 0 is also critical and reserved for padding, see {{padding}}.
```
cabo, line 595: Why is padding critical?

gselander: Good catch, should be non-critical. The processing of padding is equivalent to ignoring, which is the processing of non-critical EAD that are not understood.

draft-ietf-lake-edhoc.md
```diff
@@ -579,25 +579,40 @@ In order to reduce round trips and the number of messages, or to simplify proces
...
+
+EDHOC message_1 and the plaintext of message_2, message_3 and message_4 can be padded with the use of the corresponding EAD_x field, for x  = 1, 2, 3, 4. Padding is obtained by using an EAD item with ead_label = 0 and a random byte string of appropriate length as ead_value, noting that the ead_label and the CBOR encoding of ead_value also adds bytes. For example:
+
+* One byte padding is achieved by omitting the ead_value:
+   * EAD_x = 0x00
+* Two byte padding is achieved with the empty byte string as ead_value:
+   * ead_value = 0x40
+   * EAD_x = 0x0040
+
+Multiple occurances of EAD items with ead_label = 0 are allowed. Certain padding lengths require the use of at least two such EAD items. An EAD item with ead_label = 0 is intended to be discarded by the receiving application.
```
cabo, line 613: The second sentence should be said much earlier and certainly not in this context.

draft-ietf-lake-edhoc.md
```diff
@@ -579,25 +579,40 @@ In order to reduce round trips and the number of messages, or to simplify proces
...
 
 ~~~~~~~~~~~ CDDL
-ead = 1* (
+EAD_x = 1* ead
+
+ead = (
   ead_label : int,
-  ead_value : bstr,
+  ? ead_value : bstr,
 )
```
emanjon, line 590: ? ead_value : bstr,

Do we really want this? This is parsable but  complicates parsing. That you can not pad with 1 byte is not really an issue.

draft-ietf-lake-edhoc.md
```diff
@@ -590,9 +590,9 @@ ead = (
...
 ~~~~~~~~~~~
 
-A security application using external authorization data needs to register a positive ead_label and an associated optional ead_value format for each EAD item it uses (see {{iana-ead}}), and specify the processing and the security considerations. Each application registers its own EAD items and defines the associated operations. The application may define multiple uses of certain EAD items, e.g., the same EAD item may be used in different EDHOC messages with the same application. Multiple occurances of an EAD item in one EAD field may also be specified.
+A security application using external authorization data needs to register a ead_label and an associated optional ead_value format for each EAD item it uses (see {{iana-ead}}), and specify the processing and the security considerations. Each application registers its own EAD items and defines the associated operations. The application may define multiple uses of certain EAD items, e.g., the same EAD item may be used in different EDHOC messages with the same application. Multiple occurances of an EAD item in one EAD field may also be specified.
```
cabo, line 593:
```suggestion
A security application using external authorization data needs to register an ead_label and an associated optional ead_value format for each EAD item it uses (see {{iana-ead}}), and specify the processing and the security considerations. Each application registers its own EAD items and defines the associated operations. The application may define multiple uses of certain EAD items, e.g., the same EAD item may be used in different EDHOC messages with the same application. Multiple occurances of an EAD item in one EAD field may also be specified.
```

draft-ietf-lake-edhoc.md
cabo: Yes.  Indeed, it also provides a way to include an inexpensive "boolean" EAD, i.e., an EAD the mere presence of which already indicates useful information.

draft-ietf-lake-edhoc.md
gselander: I didn't understand the change which removed "positive". The idea was that security applications can only register positive values of ead_label, but can use the sign to indicate criticality. Two labels for the price of one. The exception is ead_label = 0, but that is anyway reserved for padding which is defined in this document. 

The current text does not make clear that negative ead_labels cannot be registered, or may lead people to think that the positive and the negative ead_label of a given absolute value may be associated to different ead_values. This was at least not the intent.

draft-ietf-lake-edhoc.md
gselander: @cabo Update in bf798e9, better now?

draft-ietf-lake-edhoc.md
gselander: There is no objection so we do this.

draft-ietf-lake-edhoc.md
```diff
@@ -579,25 +579,38 @@ In order to reduce round trips and the number of messages, or to simplify proces
...
 
-The specification registering a new EAD label needs to describe under what conditions the EAD item is critical or non-critical.
-
 The EAD fields of EDHOC must not be used for generic application data. Examples of the use of EAD are provided in {{ead-appendix}}.
 
+### Padding {#padding}
+
+EDHOC message_1 and the plaintext of message_2, message_3 and message_4 can be padded with the use of the corresponding EAD_x field, for x  = 1, 2, 3, 4. Padding is intended to be discarded by the receiving application.
+
+Padding is obtained by using an EAD item with ead_label = 0 and a random byte string of appropriate length as ead_value, noting that the ead_label and the CBOR encoding of ead_value also add bytes. For example:
```
cabo, line 605: We are not saying anything about what requirements "random" implies here.
(Clearly, the recipient cannot test on randomness 😄 )

draft-ietf-lake-edhoc.md
```diff
@@ -1796,11 +1801,16 @@ This sections compiles the CDDL definitions for ease of reference.
...
 suites = [ 2* int ] / int
 
-ead = 1* (
+ead = (
   ead_label : int,
-  ead_value : bstr,
+  ? ead_value : bstr,
 )
 
+EAD_1 = 1* ead
```
cabo, line 1809: Why do you define EAD_x and do not use it here?

draft-ietf-lake-edhoc.md
gselander: Other properties of the byte string has been discussed. @emanjon

draft-ietf-lake-edhoc.md
gselander: Ambivalence. Happy to change to

> EAD_x = 1* ead

How would CDDL handle the part:

> for x = 1, 2, 3, 4 

?

gselander:
Found an error in the CDDL (appendix C.2):

  > ? EAD_1 : ead,

draft-ietf-lake-edhoc.md
cabo: CDDL has generics for this, but in this case you just want to ignore the 1, 2, 3, 4?

```
EAD_<x> = (+ ead)
... EAD_<1> ...
```

Not sure why these gymnastics are needed.

draft-ietf-lake-edhoc.md
emanjon: No requirement on "random", but "random" is what we need for Christians use case where he wants hide information in the padding EAD without intermediaries seeing a difference.

Maybe we can write that the sender sets the value to a random string and the receiver ignores it.

draft-ietf-lake-edhoc.md
```diff
@@ -839,7 +852,7 @@ message_1 = (
   SUITES_I : suites,
   G_X : bstr,
   C_I : bstr / -24..23,
-  ? EAD_1 : ead,
+  ? EAD_1,
```
emanjon, line 855: EAD_1 is not defined, this is clearly not valid

emanjon: I suggested

  ? ead, ; EAD_1

draft-ietf-lake-edhoc.md
```diff
@@ -579,25 +579,38 @@ In order to reduce round trips and the number of messages, or to simplify proces
 
 EDHOC allows processing of external authorization data (EAD) to be defined in a separate specification, and sent in dedicated fields of the four EDHOC messages (EAD_1, EAD_2, EAD_3, EAD_4). EAD is opaque data to EDHOC.
 
-Each EAD field is a CBOR sequence (see {{CBOR}}) consisting of one or more EAD items (ead_label, ead_value) as defined below:
+Each EAD field is a CBOR sequence (see {{CBOR}}) consisting of one or more EAD items (ead_label, ? ead_value) as defined below:
 
 ~~~~~~~~~~~ CDDL
-ead = 1* (
+EAD_x = 1* ead
```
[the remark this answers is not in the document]
gselander: > EAD_1 = 1* ead
EAD_2 = 1* ead
EAD_3 = 1* ead
EAD_4 = 1* ead

was where we started 20 hours ago ;-)

Happy to leave this to @cabo and @emanjon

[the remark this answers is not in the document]
gselander: I changed my mind about leaving this. We need to merge this soon to progress on other issues.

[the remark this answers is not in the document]
gselander: > ? ead, ; EAD_1 

This doesn't work since EAD_1 etc. is one or more ead.

I reverted to having the list of EAD fields in the CDDL, see c27846a.

Unless there is something wrong with the current version, I propose we merge and then people are free to make PRs against that.

draft-ietf-lake-edhoc.md
gselander: We already say that the sender sets the value to a random byte string and the receiver ignores it. 

The word "random" which always brings questions. Is "pseudo-random" better? Then it becomes an operational condition, use output from a PRNG; rather than a question whether the byte string is truly random. 

We should also formulate to avoid questions about short paddings. The empty byte string is not random.

gselander: Proposal in 4f5b2c2

gselander:
All,

Thanks for good comments. We would like to merge this now. Any final input?

Later we can still phrase the CDDL differently, make global replace of "EAD item", "randomly generated", etc. But unless people disagree with the overall change, let's do it.

gselander:
Merging, in the absence of any quick comments.

Issue #377, opened 2022-12-09
Stephen's pre-last-call comments


gselander:
Stephen's pre-last-call comments:

https://mailarchive.ietf.org/arch/msg/lake/_5eVv4p8epYbA1yls471jqETyTU/

non-nits:

1) 3.5.2, 2nd para: on what basis is rfc5280 a normative ref
whilst [I-D.ietf-cose-cbor-encoded-cert] is informative?
They seem to be equally required for implementation based
on this text.  Given [I-D.ietf-cose-x509] is normative,
perhaps it wouldn't add delay if both were?

2) 3.5.2: I forget if use of the full DER-encoded X.509 cert
was discussed on the list or not? If not, the alternative
would have been to use the SPKI which has some slightly
different properties that can sometimes be beneficial
(potentially easier use of different CAs on each side). I'm
not asking for a change here btw, just to check was this
discussed on the list. (I see appendix D.2 talks a bit
about this.)

3) 5.2.3: what happens if the decode of message_1 fails?
(Say due to an unknown suite.) Does EAD_1 get provided to
the application?  I hope not! Maybe the text after the
bullets needs to specify this explicitly? Same for other
sections wrt providing EAD to the application (esp EAD_1
and EAD_2).

4) 9.4: As the method field is how we handle breaking
changes, do we really want that registry to be
specification required? The alternative might be standards
action, if we prefer the IESG to be involved in such
changes.

nits:

- section 1.3: this only mentions appendix A - is there a
   good reason for that or is it just text that's not
changed in ages? Probably omitting or including mention of
all appendices is better.

- section 2, p8 bullet says "Verification of the selected
   cipher suite." - that puzzled me a bit as I'd have
assumed that the transcript hashes provide/cover that
validation? If that's the case maybe this bullet could be
deleted, if not, be interested in what's meant.

- reference to [I-D.ietf-lake-traces]: not really a comment
   on edhoc, but we should probably ask the WG if we want to
hold publication of edhoc as an RFC until the traces work
is also done (i.e. create a small cluster for those two in
the RFC editor queue).

- 3.4.1, 2nd last para: the "MUST" statements here seem a
   bit bogus in 2119 interop terms. If those are only for
emphasis, then s/MUST/must/might be better to avoid other
people commenting on this later.

- 3.5.3: the two example definitions for ID_CRED_X seem to
   be introducing a notation "{ 4: kid_x }..." before that
notation is defined/referenced. Maybe add at least a
[...]

PR #378, opened 2022-12-09
Stephen's nits

gselander:
#377


sftcd:
Reading through those changes now will note things arising here, first one is below.

If an EAD item can occur more than once, do all occurrences of a given ead_label in one message or one session need to have the same criticality? Presumably if any are marked critical then all could be, and vice-versa, so not sure if mixing is needed, and it might lead to unexpected outcomes maybe?

sftcd:
Actually that's my only comment on these changes - otherwise they look good.

gselander:
@sftcd Please check fc60780

draft-ietf-lake-edhoc.md
```diff
@@ -315,15 +315,15 @@ In order to create a "full-fledged" protocol some additional protocol elements a
 
 * A keying material exporter and a key update function with forward secrecy.
 
-* Verification of the selected cipher suite.
+* Negotiation of cipher suite, more specifically verification by Responder of most preferred cipher suite by Initiator
```
emanjon, line 318: "more specifically verification by Responder of most preferred cipher suite by Initiator"

Not sure this makes thing easier to understand. This sentence is not correct, but a correct sentence would be even longer and complicated....

draft-ietf-lake-edhoc.md
```diff
@@ -529,18 +529,18 @@ ID_CRED_x may contain the authentication credential CRED_x, but for many setting
 
 EDHOC relies on COSE for identification of credentials and supports all credential types for which COSE header parameters are defined including X.509 certificates ({{I-D.ietf-cose-x509}}), C509 certificates ({{I-D.ietf-cose-cbor-encoded-cert}}), CWT ({{cwt-header-param}}) and CWT Claims Set (CCS) ({{cwt-header-param}}).
 
-ID_CRED_I and ID_CRED_R are COSE header maps and contains one or more COSE header parameters. ID_CRED_I and ID_CRED_R MAY contain different header parameters. The header parameters typically provide some information about the format of the credential.
+ID_CRED_I and ID_CRED_R are of type COSE header_map, as defined in Section 3 of {{RFC9052}}, and contains one or more COSE header parameters. ID_CRED_I and ID_CRED_R MAY contain different header parameters. The header parameters typically provide some information about the format of the credential.
```
emanjon, line 532: That is good!

draft-ietf-lake-edhoc.md
```diff
@@ -529,18 +529,18 @@ ID_CRED_x may contain the authentication credential CRED_x, but for many setting
...
 
-Note that COSE header parameters in ID_CRED_x are used to identify the message sender's credential. There is therefore no reason to use the "-sender" header parameters, such as x5t-sender, defined in Section 3 of {{I-D.ietf-cose-x509}}. Instead, the corresponding parameter without "-sender", such as x5t, SHOULD be used.
-
-Example: X.509 certificates can be identified by a hash value using the 'x5t' parameter:
+Example: X.509 certificates can be identified by a hash value using the 'x5t' parameter, see Section 2 of {{I-D.ietf-cose-x509}}:
```
emanjon, line 534: This seems bad. Why? EDHOC is using COSE header_map but changing the definition of some of the parameters. Why should that not be described?

emanjon (approved):

draft-ietf-lake-edhoc.md
gselander: Abbreviated, see 67947ff below.

Issue #380, opened 2022-12-09
Stephen's non-nits


gselander:
From #377 

non-nits:

    3.5.2, 2nd para: on what basis is rfc5280 a normative ref
    whilst [I-D.ietf-cose-cbor-encoded-cert] is informative?
    They seem to be equally required for implementation based
    on this text. Given [I-D.ietf-cose-x509] is normative,
    perhaps it wouldn't add delay if both were?

I made RFC 5820 informative.

    3.5.2: I forget if use of the full DER-encoded X.509 cert
    was discussed on the list or not? If not, the alternative
    would have been to use the SPKI which has some slightly
    different properties that can sometimes be beneficial
    (potentially easier use of different CAs on each side). I'm
    not asking for a change here btw, just to check was this
    discussed on the list. (I see appendix D.2 talks a bit
    about this.)

The use of (full DER-encoded) X.509 is described and planned to be used in deployments. I think Marek mentioned that in his presentation at IETF 113. Should be known. I don't think SPKI has been discussed. Any action you propose?

    5.2.3: what happens if the decode of message_1 fails?
    (Say due to an unknown suite.) Does EAD_1 get provided to
    the application? I hope not! Maybe the text after the
    bullets needs to specify this explicitly? Same for other
    sections wrt providing EAD to the application (esp EAD_1
    and EAD_2).

The current text says:
> The Responder SHALL process message_1 as follows:
>
> * Decode message_1 (see [Appendix C.1]
> * Verify that the selected cipher suite is supported and that no prior cipher suite in SUITES_I is supported.
> * If EAD_1 is present then make it available to the application for EAD processing.
>
> If any processing step fails, then the Responder MUST send an EDHOC error message back as defined in [Section 6], and the protocol MUST be discontinued.

The intent here and all other message processing is that you carry out the processing in steps and if any step fail then you don't carry out the subsequent steps. Is that clear enough? Then that would cover the case of unknown suite. But perhaps it doesn't cover other errors occuring at reception of message_1.

Does the state machine #373 make this more clear (though just an example)? Should we move some of that text here?

    9.4: As the method field is how we handle breaking
    changes, do we really want that registry to be
    specification required? The alternative might be standards
    action, if we prefer the IESG to be involved in such
    changes.

This has been discussed. One option is to require standards action for low method numbers. Anyone else has an opinion?

sftcd:
> I made RFC 5820 informative.

[...]

PR #381, opened 2022-12-09
Stephen's non-nits

gselander:
#380


gselander:
This is a PR on top of the Stephen's-nits PR. The recent commit addressed this non-nit:

>> The intent here and all other message processing is that you carry out the processing in steps and if any step fail then you don't carry out the subsequent steps. Is that clear enough?
>
> I'm not sure it's clear enough - esp wrt passing on EAD which can have side-effects that could be bad (e.g. if some identifier were stored even though message processing had failed for some reason). I think the EAD stuff is the only way that kind of thing might happen - if so, then maybe saying that EAD should only be passed to applications after all processing steps have succeeded might be right?

gselander:
@sftcd Please check 7631715

emanjon:
This is hard to review due to the overlap with the nits PR. The nits PR seems ready to merge, this is not.

emanjon:
I looked at the commits 89b46cc and 5d6913d

They look good and ready to merge

gselander:
Closing this in favor of #383

PR #383, opened 2022-12-19
Stephens non nits rebased

gselander:
This PR is the same as #381, but rebased on the current master.


emanjon (approved):

gselander:
Merging this now to have an updated master to work on for #275 and #384

Issue #396, opened 2023-01-27
Security of using the same key for Initiator and Responder roles


emanjon:
The document writes:

```
As noted in Section 12 of [RFC9052] the use of a single key for
   multiple algorithms is strongly discouraged unless proven secure by a
   dedicated cryptographic analysis.  In particular this recommendation
   applies to using the same private key for static Diffie-Hellman
   authentication and digital signature authentication.  A preliminary
   conjecture is that a minor change to EDHOC may be sufficient to fit
   the analysis of secure shared signature and ECDH key usage in
   [Degabriele11] and [Thormarker21].
```

What can we say about using the same key for Initiator and Responder roles? We should probably say something

charlie-j:
In principles, a signature for the initiator role cannot be confused with a responder's one (it would require crazy hash and mac collision I guess), so this does not seem to pose any particular problem. Our models did include this possibility, by default a key could be used for both roles, which gives some additional confidence toward this intuition.

Did you have any particular threat in mind?

emanjon:
>a signature for the initiator role cannot be confused with a responder's one 
>by default a key could be used for both roles, which gives some additional confidence toward this intuition.

Thanks. I will add text that the same key for Initiator and Responder roles.

>Did you have any particular threat in mind?
No. Just if a signature for the initiator role can be confused with a responder's one.

charlie-j:
One additional thought on this -> if we want to completely rule out any potential issue here, we can add inside the signature payload a constant flag, with two distinct values for the responder and initiator side.

That simplifies the overall crypto argument for the security, potentially at no cost?

gselander:
We discussed this among the authors. This is a good idea, but we wouldn't like to change the test vectors at this late stage unless we think it is necessary. Let's bring it up for discussion if we for some other reason need to update the test vectors. Closing this now.

PR #397, opened 2023-02-27
#396 - using the same key for Initiator and Responder

emanjon:
#396 - using the same key for Initiator and Responder


emanjon:
The document need to say something about this.
It seems very uncontroversial to state that it MAY be used for both.
I think this can be merged and the issue closed.

gselander (commented):
Proposal: 
OLD
"The same authentication key MAY be used for both the Initiator and Responder roles."
NEW
"The same authentication credential MAY be used for both the Initiator and Responder roles."

Issue #375, opened 2022-12-07
Protected error, EDHOC-exporter, and responder processing of message_3


emanjon:
See

After sending message_3 the Initiator can use the exporter. When the responder receives message_3 the message might parse perfectly and the responder can use the exporter. The responder might still not want to talk to the Initiator. In general it makes sense that the Responder can use the derived application keys to protect the EDHOC error message. This corresponds to protected alert messages in e.g., TLS.

As far as I can see -17 does not say anything about this. It propably should say a sentence or two about this.

Releted issues.
https://github.com/core-wg/oscore-edhoc/issues/8
https://github.com/lake-wg/edhoc/issues/372

gselander:
Current processing of message_3:

> Verify Signature_or_MAC_3 using the algorithm in the selected cipher suite. The verification process depends on the method, see [Section 5.4.2]. Make the result of the verification available to the application.

> After verifying message_3, the Responder can compute PRK_out, see Section 4.1.3, derive application keys using the EDHOC-Exporter interface, see Section 4.2.1, persistently store the keying material, and send protected application data.

The application is in charge of using or not using the EDHOC session, this seems out of scope of EDHOC?

emanjon:
```
* Make ID_CRED_R and (if present) EAD_2 available to the application for authentication- and EAD processing.

* Obtain the authentication credential (CRED_R) and the authentication key of R from the application (or by other means).

* Verify Signature_or_MAC_2 using the algorithm in the selected cipher suite. The verification process depends on the method, see [Section 5.3.2](https://datatracker.ietf.org/doc/html/draft-ietf-lake-edhoc#asym-msg2-proc). Make the result of the verification available to the application.

If any processing step fails, then the Initiator MUST send an EDHOC error message back as defined in [Section 6](https://datatracker.ietf.org/doc/html/draft-ietf-lake-edhoc#error), and the protocol MUST be discontinued.
```
Some thought on the processing

Do we want as strong focus on the the authentication credential is coming from the application? Maybe these would be improvements.

* Obtain the authentication credential (CRED_R) and the authentication key of R.
* Obtain the authentication credential (CRED_R) and the authentication key of R (from the application or by other means).

[...]

PR #398, opened 2023-02-27
#375 processing of message_2 and message_3

emanjon:
#375 processing of message_2 and message_3


emanjon:
I think these sections need more considerations and clarifications. Some people has stated that you cannot derive OSCORE keys if you get an EDHOC error. My understanding was that you could. There are two types of EDHOC errors. One parsing errors to you cannot derive keys and then authentication errors (X.509 identity not authorized, X.509 cert expired, X.509 issuer not trusted, OCSP server offline, etc.). Whatever the answer is it needs to be clear. The current situation where people have different understandings is not good.

With the first type of error, TLS sends an unprotected error. In the second case TLS sends an protected error.

gselander:
Perhaps we should break out the error discussion in a separate issue? 

Just as a reminder, at IETF 110 we had the following candidate EDHOC error classes:

1. Selected cipher suite not supported
2. Message (field) error (syntax error or incorrect protocol field of expected message)
3. Credential/context error (error related to establish certificate/raw public key context for processing message, including certificate expired, revoked, unsupported, corrupt, unknown CA, or other issue in processing the credential)
4. Decrypt error (a cryptographic operation failed, including being unable to correctly verify a signature or a MAC in the protocol message)
5. EAD error (unsupported or missing data)
6. Access denied (credential valid but peer not allowed access)
7. Internal error (error not related to the protocol or the peer)

- 1 is included in the draft
- 3 is related to the authentication errors John mentioned above
- 5 has been brought up again, with a registry for EAD applications to register their own errors
- a new error code is proposed in #401

emanjon:
> Perhaps we should break out the error discussion in a separate issue?
> 
> Just as a reminder, at IETF 110 we had the following candidate EDHOC error classes:
> 
>     1. Selected cipher suite not supported
> 
>     2. Message (field) error (syntax error or incorrect protocol field of expected message)
> 
>     3. Credential/context error (error related to establish certificate/raw public key context for processing message, including certificate expired, revoked, unsupported, corrupt, unknown CA, or other issue in processing the credential)
> 
>     4. Decrypt error (a cryptographic operation failed, including being unable to correctly verify a signature or a MAC in the protocol message)
> 
>     5. EAD error (unsupported or missing data)
> 
>     6. Access denied (credential valid but peer not allowed access)
> 
>     7. Internal error (error not related to the protocol or the peer)
> 
> 
>     * 1 is included in the draft
> 
>     * 3 is related to the authentication errors John mentioned above
> 
>     * 5 has been brought up again, with a registry for EAD applications to register their own errors
> 
>     * a new error code is proposed in [New error code for unknown credential referenced #401](https://github.com/lake-wg/edhoc/pull/401)

Good to take that as part of the #401 discussion. Currently 2-7 are in the draft as "unspecified error".

gselander:
The two points we have been discussing are:

* When does authentication occur?
* When can the keys derived from PRK_out be used?

@marco-tiloca-sics @chrysn Do you have any objections to this PR?

gselander:
Note that the current text seems to assume that error messages are not secured:

OLD
> But as the error message is not authenticated, a received error message might also have been sent by an attacker and the receiver MAY therefore try to continue the protocol.

This text indicates that a derived key cannot be used for protecting the error. Needs to be clarified, e.g. rephrase sentence above as:
NEW
But as the error message is expected to not be authenticated, a received error message might also have been sent by an attacker and the receiver MAY therefore try to continue the protocol.

marco-tiloca-sics:
This PR looks good to me.

Also, the clarification proposed in the comment above [1] looks good.

[1] https://github.com/lake-wg/edhoc/pull/398#issuecomment-1580322669

gselander:
Merging this as no objection was received.

PR #399, opened 2023-02-28
Editorial EDHOC_KDF changes

emanjon:


emanjon:
I strongly think that all 10 EDHOC-KDF should be in the same table
The text about mac_length_x need to use the defined term hash_length

gselander (approved):
I made one commit for review.

PR #406, opened 2023-05-12
Add example to clarify explanation on Wrong Selected Cipher Suite

geonnave:
Adding an example to clarify the explanation on cipher suite negotiation.

The reason I think this could help is that:
- in `(...) if the selected cipher suite is the first cipher suite in SUITES_I that the Responder supports (...)`
- the part `first cipher suite` can be confusing with the fact that SUITES_I is an array whose first item is the most preferred cipher suite by the initiator.
- however, in this case, `first cipher suite` does not refer to the first item in the list SUITES_I, but rather to the first item that is supported by the Responder.
- the idea of adding the example is to clarify that.


gselander (commented):
Good catch, this part is not exemplified. I propose to keep the examples together in the following section. Please review  44afd42.

geonnave:
Looks good to me!

PR #410, opened 2023-06-13
Encrypting C_R

emanjon:


gselander (approved):

emanjon:
@marco-tiloca-sics Do you have any opinions on the position of C_R inside the ciphertext?

charlie-j:
Just say this PR after sending a mail on the ML, but some comments on this:

Hi all,

Seeing this a bit late, but note that this implies to change the TH_2 computation:

TH_2 contains C_R, and TH_2 is used to compute  KEYSTREAM_2, itself finally used to encrypt the content of message PLAINTEXT_2.

So, if we put C_R in the PLAINTEXT_2, it needs to be dropped from TH_2, and then added to TH_3.

As a side a node, including C_R in the plaintext will also imply that the construction from https://datatracker.ietf.org/doc/html/draft-ietf-lake-edhoc-19#name-long-plaintext_2 will also need to be used more frequently, which is debatable whether it is a good thing.

Best,

Charlie

emanjon:
If we encrypt. @marco-tiloca-sics  would like to have C_R first in the plaintext

marco-tiloca-sics:
> If we encrypt. @marco-tiloca-sics would like to have C_R first in the plaintext

Yes, that's in order to be kind with strictly sequential parsers that can find C_R at the beginning of PLAINTEXT_2, i.e., as soon as possible.

chrysn:
The error handling aspect doesn't seem all that bad to me. In the typical scenario with a client as initiator, it's the client that has to act on the error anyway (eg. by sending different EAD1 or whatever else can be done in that situation). In the reverse direction with the server as Initiator, the client can still send M2 easily (because C_I is unencrypted), and on error processing M2, M3 reaches the client on the return channel without explicit use of C_R.

gselander:
About the side note by @charlie-j on increased size of PLAINTEXT_2 and the need to use the Long PLAINTEXT_2 mechanism more frequently: I don't think this is a major drawback since the cut-off is larger than 8 kB and the size of C_R typically much smaller.

gselander:
As there is WG consensus to encrypt C_R
https://mailarchive.ietf.org/arch/msg/lake/W9tEMIhGT_i5SzbjrR9BNy_Biqs/
we plan to merge this PR soon to make the draft ready for last call. Further comments are still welcome.

Issue #403, opened 2023-03-29
kccs and kcwt security considerations.


emanjon:
The document registers two new COSE header parameters that can be used for COSE in general (not only EDHOC) but does not give any security considerations for them. As a comparision, the COSE X509 has quite a lot of things to say for those 4 new header parameters.

Security considerations are explicitly important for kccs that transport a new public key but not any information to verify that the key is trusted. That kind of information need to be come from elsewhere. If not the best security is TOFU. This is quite different from kcwt and x5t, kid, x5chain, etc.

emanjon:
https://datatracker.ietf.org/doc/rfc9360/ gives an example of the type of considerations that might be needed

PR #411, opened 2023-07-07
CWT and CCS security considerations

emanjon:
#403


gselander:
Merging now to prepare for -20

Issue #412, opened 2023-07-07
or a pseudorandom number such as a hash


emanjon:
A hash is deterministic and not pseudorandom. A digest is only pseudorandom is the input is pseudorandom.

emanjon:
"e.g., be a counter or a pseudorandom number such as a hash"

PR #413, opened 2023-07-07
Correction addressing #412

gselander:


gselander:
Merging to prepare for -20

PR #434, opened 2023-08-25
Update 3.4

gselander:
Mentioning preference for reliable transport

### DIFF
--- a/draft-ietf-lake-edhoc.md
+++ b/draft-ietf-lake-edhoc.md
@@ -498,7 +498,7 @@ Examples:
 
 Cryptographically, EDHOC does not put requirements on the underlying layers. Received messages are processed as the expected next message according to protocol state, see {{asym}}. If processing fails for any reason then, typically, an error message is attempted to be sent and the EDHOC session is aborted.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-lake-edhoc.md
```diff
@@ -498,7 +498,7 @@ Examples:
 
 Cryptographically, EDHOC does not put requirements on the underlying layers. Received messages are processed as the expected next message according to protocol state, see {{asym}}. If processing fails for any reason then, typically, an error message is attempted to be sent and the EDHOC session is aborted.
 
-EDHOC is not bound to a particular transport layer and can even be used in environments without IP. Ultimately, the application is free to choose how to transport EDHOC messages including errors. However, EDHOC was designed with CoAP in reliable mode as default transport. In order to avoid unnecessary message processing or protocol termination, the transport is responsible to handle:
+EDHOC is not bound to a particular transport layer and can even be used in environments without IP. Ultimately, the application is free to choose how to transport EDHOC messages including errors. In order to avoid unnecessary message processing or protocol termination, it is preferred to use reliable transport such as CoAP in reliable mode, which is the default transport. The transport is responsible to handle:
```
zaheduzzaman: This is getting close to where I think it should be. I would suggest to rephrase like following -

EDHOC is not bound to a particular transport layer and can even be used in environments without IP. Ultimately, the application is free to choose how to transport EDHOC messages including errors. In order to avoid unnecessary message processing at the application or protocol termination, it is RECOMMENDED to use a reliable transport ( EDHOC originally designed to be transported over CoAP in reliable mode, see appendix A). The transport, in general, SHOULD handle:

zaheduzzaman (approved):
LGTM, thanks.

emanjon (approved):

PR #442, opened 2023-08-30
Traces: Examples of invalid EDHOC messages

emanjon:


draft-ietf-lake-traces.md
```diff
@@ -2817,6 +2838,73 @@ OSCORE Master Salt after KeyUpdate (Raw Value) (8 bytes)
...
+~~~~~~~~
+Invalid message_1 (38 bytes)
+03 06 58 20 74 1a 13 d7 ba 04 8f bb 61 5e 94 38 6a a3 b6 1b ea 5b 3d
+8f 65 f3 26 20 b7 49 be e8 d2 78 ef a9 41 0e
+~~~~~~~~
+
+Invalid number of elements in the CBOR sequence.
+
+~~~~~~~~
+Invalid message_1 (38 bytes)
```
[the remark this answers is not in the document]
marco-tiloca-sics: Yes, I agree.

The only exception would be if this particular EAD item is defined to have a mandatory value, but for that we'd need such an EAD item to be specified and registered already.

malishav:
These are some detailed tests on parsing message_1 and validating the point. The question is why would we do this only for message_1 and not for e.g. message_2 as well? I agree that these are very useful and I already caught several bugs in my implementation that need to be fixed for these tests to pass.

draft-ietf-lake-traces.md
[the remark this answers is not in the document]
emanjon: Good, I removed it for now as it we already have a "wrong type" example.

emanjon:
>I agree that these are very useful and I already caught several bugs in my implementation that need to be fixed for these tests to pass.

That is a very good argument for why we should include this.

>The question is why would we do this only for message_1 and not for e.g. message_2 as well?

As a first start I just came up with as many different types of invalidity as possible, trying to not repeat things. My idea was that the implementer would need to make sure that the same type of invalidity is checked in other parts as well. That should probably be written out.

Yes, I think it is a good idea to have some invalid examples for message_2 as well. There we could have a "wrong number of items" message_2 as well as PLAINTEXT_2 with invalid ID_CRED_R.

emanjon:
@malishav I added some invalid message_2 and PLAINTEXT_2 and added some more text in the beginning of the section. Please check

emanjon:
> This message indicates cipher suite 6, which has X25519 as EDHOC key exchange algorithm (ECDH curve).
> 
> Since the error is about G_X not corresponding to a point on the P-256 curve, it is better for consistency to indicate cipher suite 2, 3 or 5 instead.
> 

Indeed. I just used the example in 4.1. message_1 (first time) and missed that 6 is not P-256. I'll change to 2

draft-ietf-lake-traces.md
```diff
@@ -2817,6 +2838,94 @@ OSCORE Master Salt after KeyUpdate (Raw Value) (8 bytes)
...
+00 ff ff ff ff ff ff ff ff ff ff ff ff 0e
+~~~~~~~~
+
+Invalid x-coordinate in (G_X) not corresponding to a point on the P-256 curve.
+
+~~~~~~~~
+Invalid message_1 (37 bytes)
+03 06 58 20 a0 4e 73 60 1d f5 44 a7 0b a7 ea 1e 57 03 0f 7d 4b 4e b7
+f6 73 92 4e 58 d5 4c a7 7a 5e 7d 4d 4a 0e
+~~~~~~~~
```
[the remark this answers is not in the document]
emanjon: fixed

draft-ietf-lake-traces.md
```diff
@@ -2817,6 +2838,94 @@ OSCORE Master Salt after KeyUpdate (Raw Value) (8 bytes)
...
+58 20 41 97 01 d7 f0 0a 26 c2 dc 58 7a 36 dd 75 25 49 f3 37 63 c8 93
+42 2c 8e a0 f9 55 a1 3a 4f f5 d5 4B 98 62 a1 1d e4 2a 95 d7 85 38 6a
+~~~~~~~~
+
+Invalid encoding of ID_CRED_R in PLAINTEXT_2
+
+~~~~~~~~
+Invalid  PLAINTEXT_2 (15 bytes)
+27 a1 04 42 32 10 48 fa 5e fa 2e bf 92 0b f3
+~~~~~~~~
```
marco-tiloca-sics, line 2914: (This comment was made on a commit, but I add it here too to be sure it's not missed)



What is the invalidity in the new example?

This seems the correct way to encode the credential identifier of type 'kid' with value 0x3210.

emanjon: My understanding of Section 3.5.3.2. is that the CBOR map { 4 : kid_x } is always replaced by the byte string kid_x.

I.e. the correct encoding of this PLAINTEXT_2 is 
27 42 32 10 48 fa 5e fa 2e bf 92 0b f3

marco-tiloca-sics: Ops, you're right :-)

Maybe it's useful if these examples shortly mention what their invalidity is, possibly pointing to the relevant section of -lake-edhoc

emanjon: That is a good idea

emanjon: Done

emanjon:
Added that implementations must not compose these examples according to EDHOC. Added reference to RFC 8949

Added more invalid examples concerning SUITES_I and ID_CRED_R.

Please review.

draft-ietf-lake-traces.md
```diff
@@ -2817,6 +2838,118 @@ OSCORE Master Salt after KeyUpdate (Raw Value) (8 bytes)
...
+Invalid encoding of message_1 as array. Correct encoding is a CBOR sequence according to Section 5.2.1 of {{I-D.ietf-lake-edhoc}}.
+
+~~~~~~~~
+Invalid message_1 (38 bytes)
+84 03 02 58 20 74 1a 13 d7 ba 04 8f bb 61 5e 94 38 6a a3 b6 1b ea 5b
+3d 8f 65 f3 26 20 b7 49 be e8 d2 78 ef a9 0e
+~~~~~~~~
+
+Invalid 16-bit encoding 19 00 03 of METHOD = 3. Correct is the deterministic encoding 03 according to Section 1.4 of {{I-D.ietf-lake-edhoc}} and Section 4.2.1 of {{RFC8949}}.
+
```
marco-tiloca-sics, line 2854: I think it is better to refer to Section 3.1 of -lake-edhoc (instead of Section 1.4).

That's where deterministic encoding is mentioned as a property of the protocol (emphasis mine):

> All **EDHOC messages** are CBOR Sequences [[RFC8742](https://www.rfc-editor.org/info/rfc8742)], and **are deterministically encoded**.

while Section 1.4 is more about terminology and language for the document (emphasis mine):

> When referring to CBOR, this **specification** always **refers** to Deterministically Encoded CBOR as specified in Sections 4.2.1 and 4.2.2 of [[RFC8949](https://www.rfc-editor.org/info/rfc8949)].

draft-ietf-lake-traces.md
```diff
@@ -2817,6 +2838,118 @@ OSCORE Master Salt after KeyUpdate (Raw Value) (8 bytes)
...
+Invalid encoding 41 32 of ID_CRED_R in PLAINTEXT_2. Correct encoding is 32 according to Section 3.5.3.2 of {{I-D.ietf-lake-edhoc}}.
+
+~~~~~~~~
+Invalid PLAINTEXT_2 (12 bytes)
+27 41 32 48 fa 5e fa 2e bf 92 0b f3
+~~~~~~~~
+
+Invalid encoding of the map ID_CRED_R in PLAINTEXT_2. Correct is the deterministic bytewise lexicographic order encoding according to Section 1.4 of {{I-D.ietf-lake-edhoc}} and Section 4.2.1 of {{RFC8949}}.
+
```
marco-tiloca-sics, line 2940: Instead of Section 1.4 of -lake-edhoc, I think it is better to refer to both Sections 3.1 (as mentioning deterministic encoding) and Section 3.5.3 (as talking about ID_CRED_X and its encoding).

Note that Section 3.5.3 is not *explicitly* saying that the CBOR map encoding ID_CRED_X has to be deterministically encoded; it is just silently inheriting that requirement from Section 3.1.

Is it worth adding an explicit reminder in Section 3.5.3 in -lake-edhoc? Such a reminder is in fact present in Section 3.5.2 when discussing authentication credentials instead.

draft-ietf-lake-traces.md
```diff
@@ -2817,6 +2838,118 @@ OSCORE Master Salt after KeyUpdate (Raw Value) (8 bytes)
...
+Invalid encoding 41 0e of C_I = 0x0e. Correct encoding is 0e according to Section 3.3.2 of {{I-D.ietf-lake-edhoc}}.
+
+~~~~~~~~
+Invalid message_1 (38 bytes)
+03 02 58 20 74 1a 13 d7 ba 04 8f bb 61 5e 94 38 6a a3 b6 1b ea 5b 3d
+8f 65 f3 26 20 b7 49 be e8 d2 78 ef a9 41 0e
+~~~~~~~~
+
+Invalid array encoding 81 02 of SUITES_I = 2. Correct encoding is 02 according to Section 5.2.1 of {{I-D.ietf-lake-edhoc}}.
+
```
marco-tiloca-sics, line 2870: i think it is more precise to refer to Section 5.2.2 of -lake-edhoc; that's where the two variants for encoding SUITES_I are defined.

emanjon: Can do

draft-ietf-lake-traces.md
emanjon: Done. I will try to update the EDHOC specification during AUTH48 so that section 3.1 refers to 4.2.1 of [[RFC8949](https://www.rfc-editor.org/info/rfc8949)].

draft-ietf-lake-traces.md
emanjon: I will suggst to add an explicit reminder in Section 3.5.3 during AUTH48

emanjon:
If we want to have heading I think they should describe the general type of error instead of the specific example.

emanjon:
Maybe some headings
- Invalid G_X (cddl correct)
- Invalid cddl
- Nor deterministic encoding

and then subbullets inside

Sent from Outlook for iOS<https://aka.ms/o0ukef>
________________________________

gselander:
The third category - non-deterministic CBOR - is not necessarily invalid. See discussion in #446.

emanjon:
Added a curve25519 point (p) of low order which produces the all zero value
https://github.com/jedisct1/libsodium/blob/master/src/libsodium/crypto_scalarmult/curve25519/ref10/x25519_ref10.c#L17

PR #446, opened 2023-09-04
EDHOC: Encoding clarifications

emanjon:
Intent is to provide clarifications regarding encoding. Suggested from implementors that this is needed.

The PR is not finished. There will be more commits.

### DIFF
--- a/draft-ietf-lake-edhoc.md
+++ b/draft-ietf-lake-edhoc.md
@@ -392,7 +392,7 @@ To simplify for implementors, the use of CBOR and COSE in EDHOC is summarized in
 
 The EDHOC protocol consists of three mandatory messages (message_1, message_2, message_3), an optional fourth message (message_4), and an error message, between an Initiator (I) and a Responder (R). The odd messages are sent by I, the even by R. Both I and R can send error messages.
 The roles have slightly different security properties which should be considered when the roles are assigned, see {{sec-prop}}.
-All EDHOC messages are CBOR Sequences {{RFC8742}}, and are encoded using deterministically encoded CBOR as specified in Section 4.2.1 of {{RFC8949}}. {{fig-flow}} illustrates an EDHOC message flow with the optional fourth message as well as the content of each message. The protocol elements in the figure are introduced in {{overview}} and {{asym}}. Message formatting and processing are specified in {{asym}} and {{error}}.
+All EDHOC messages are CBOR Sequences {{RFC8742}}, and are defined to be deterministically encoded CBOR as specified in Section 4.2.1 of {{RFC8949}}. {{fig-flow}} illustrates an EDHOC message flow with the optional fourth message as well as the content of each message. The protocol elements in the figure are introduced in {{overview}} and {{asym}}. Message formatting and processing are specified in {{asym}} and {{error}}.
 
 Application data may be protected using the agreed application algorithms (AEAD, hash) in the selected cipher suite (see {{cs}}) and the application can make use of the established connection identifiers C_I and C_R (see {{ci}}). Media types that may be used for EDHOC are defined in {{media-type}}.
 


gselander:
The PR is now changing from "RECOMMENDED to abort" to "MAY abort" because it is thought to simplify implementation and there are no obvious security reasons to prevent non-deterministic CBOR.

@marco-tiloca-sics @malishav @chrysn It would be good to get some feedback on the difficulty/ease of enforcing deterministic CBOR.

marco-tiloca-sics:
> The PR is now changing from "RECOMMENDED to abort" to "MAY abort" because it is thought to simplify implementation and there are no obvious security reasons to prevent non-deterministic CBOR.
> 
> @marco-tiloca-sics @malishav @chrysn It would be good to get some feedback on the difficulty/ease of enforcing deterministic CBOR.

At least the CBOR library that I use does not seem to provide a method for checking that incoming CBOR data is deterministically encoded.

Implementing those checks does not feel like challenging (I'm doing that for CBOR integers at the moment), but avoiding the need for having those in place certainly simplifies things.

I think the advantage is also functional when thinking of authentication credentials (which was my original point about this topic). If CRED_X is CBOR based (e.g., a CCS or a CWT) and it is transported by value in ID_CRED_X, it's great that the recipient peer can use CRED_X as is, i.e., even when it is not deterministically encoded.

However, since the case above has CRED_X specified by value in ID_CRED_X, which is a field of PLAINTEXT, then forcing deterministic encoding of messages would also imply deterministic encoding of CBOR based credentials that are transported by value in ID_CRED_X (so that the requirement on the message as a whole is fulfilled).

If non-deterministic encoding can be admitted while still being safe in terms of security, that it feels better to not be strict in mandating it (while a peer might still be strict about it and MAY abort the session anyway as per the new phrasing).

emanjon:
EDHOC-22 states:

   The Initiator
   and Responder SHOULD use an available authentication credential
   (transported in EDHOC or otherwise provisioned) without re-encoding.

My and Görans interpretation is that the byte string value of bccs and kcwt are treated as opaque strings by EDHOC when creating the messages. I.e. no requirement to encode deterministic. Let's clarify this.

emanjon:
> At least the CBOR library that I use does not seem to provide a method for checking that incoming CBOR data is deterministically encoded.
> 
> Implementing those checks does not feel like challenging (I'm doing that for CBOR integers at the moment), but avoiding the need for having those in place certainly simplifies things.

That is good input. If I remember correctly, deterministic encoding of the fields before G_X in message one had a security impact if the hash function (SHA-2) turns out to have significant weaknesses. 

   message_1 = (
     METHOD : int,
     SUITES_I : suites,
     G_X : bstr,
     C_I : bstr / -24..23,
     ? EAD_1,
   )

Should maybe clarify that.

chrysn:
The libraries I know off my head don't provide checking; sure it can be performed, but without library modifications it'd be re-encoding, and that's tedious (eg. because you need an extra buffer, or you need to be careful about overwriting in-place). Unless there are concrete security concerns, I'd prefer requirement on verifying deterministic rules (but I'm fine with keeping the requirement that things be encoded deterministically).

There are two arguments I see against it (divergent decoding may be overlooked when testing against relaxed peers; and the covert channel), but both have good counterarguments (read the specs and test with more than one peer; there are already optional EADs and there's nobody to hide communication from because by EDHOC design only the peers can read the plaintext).

emanjon:
Approval of the PR would mean that the following invalid test vector should be removed from draft-ietf-lake-traces

> 
>    4.3.3.  Non-lexicographic order of map
> 
>    Invalid encoding of the map ID_CRED_R in PLAINTEXT_2.  Correct is the
>    deterministic bytewise lexicographic order encoding according to
>    Section 3.1 of [I-D.ietf-lake-edhoc] and Section 4.2.1 of [RFC8949].
> 
>    Invalid PLAINTEXT_2 (25 bytes)
>    27 A2 0A 46 6D 6F 72 64 6F 72 04 44 72 69 6E 67 48 fa 5e fa 2e bf 92
>    0b f3
>

malishav:
Is this ready for merge? Let me know if I should poll the WG for final comments?

gselander (approved):

gselander:
I just responded to the list.

emanjon:
This PR #446 has been sent to the RFC editor.

gselander:
Merging for submission of -23.

